PR #4, opened 2011-12-15
Externalize strings

Pike:
externalize localizable strings to a DTD, reference those from the xml.in
and pick up a DOCTYPE from the template


Pike:
https://github.com/mozilla-services/android-sync/commit/55e0dc7260db37eafec39ce419d50175dd83c5b3 is effectively this, thanks.

Closing the request.

PR #8, opened 2012-01-09
Remove cast warnings.

emtwo:
In addition to the redundant cast warning changes, I found that the unchecked warnings shown here

https://bugzilla.mozilla.org/show_bug.cgi?id=716073#c0

can be resolved by parameterizing the JSONObject class definition as shown here

http://pastebin.mozilla.org/1440881

based on the issue from here

http://code.google.com/p/json-simple/issues/detail?id=15

However, I'm not sure if it's okay to modify this since it's an external dependency.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 AndroidManifest.xml
+R.java
 src/main/java/org/mozilla/gecko/sync/repositories/android/Authorities.java
 res/xml/sync_syncadapter.xml
 res/values/strings.xml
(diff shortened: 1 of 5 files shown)


rnewman:
> However, I'm not sure if it's okay to modify this since it's an external dependency.

We could probably get away with it, but I'd rather not spend time or attention on that. We're hoping to move to Jackson as soon as it's convenient to do so.

Good investigation, though!

PR #13, opened 2012-01-10
Error handling + no network connectivity handling + style cleanup.

liuche:


liuche:
Added alignment in Constants (I only added the flags constant!); unfortunately, REORDER_TO_FRONT does not work as well as one would hope: if a screen (e.g., Android's settings) is already open, it does not get reordered to front, so going there pushes another Android settings on the stack. Don't know if that can be fixed.

rnewman:
Urgh, Android's UI stuff sucks.

PR #26, opened 2012-01-14
Move "Pair Device" into Firefox Sync settings, per design review.

liuche:
Show one-account-supported alert when attempting to add more than one Firefox Sync account. Pair Device can be accessed from Firefox Sync settings in Sync account menu. Quick fixes for VKB problems. Added some strings (will file localization bug next).


rnewman:
Don't bother with a strings bug.

PR #42, opened 2012-01-27
Bug 709348 - equality work.

rnewman:
These patches introduce a layered definition of equality between records, allowing us to write more sane tests and reconciling code.

These patches don't change the existing reconciling code (that will come shortly), but they do expose and fix a data-loss bug when handling deleted records. Test code wasn't checking the visits of a history record, which meant that the data loss was ignored.

Hooray for sane primitives!


rnewman:
Tests to follow.

PR #43, opened 2012-01-31
Bug 709348 - rework storing of records, track stored records to prevent reupload

rnewman:
This hasn't been tested on a real device, but it does pass automated tests (as well as a new one), and broke tests in the expected places :)

Input requested. Note the long sequence of commits that includes some evolutionary steps for clarity.

Nick, you'll notice a bunch of commits here for moving WBORepository. I hope that doesn't spike you too much!

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/repositories/HashSetStoreTracker.java
+++ b/src/main/java/org/mozilla/gecko/sync/repositories/HashSetStoreTracker.java
@@ -6,6 +6,8 @@ package org.mozilla.gecko.sync.repositories;
 
 import java.util.HashSet;
 
+import org.mozilla.gecko.sync.repositories.domain.Record;
+
 public class HashSetStoreTracker implements StoreTracker {
 
   // Guarded by `this`.
@@ -35,5 +37,18 @@ public class HashSetStoreTracker implements StoreTracker {
   @Override
   public synchronized boolean untrackStoredForExclusion(String guid) {
     return (guid != null) && guids.remove(guid);
+  }
+
+  @Override
+  public RecordFilter getFilter() {
+    if (guids.size() == 0) {
+      return null;
+    }
+    return new RecordFilter() {
+      @Override
+      public boolean excludeRecord(Record r) {
+        return isTrackedForExclusion(r.guid);
+      }
+    };
   }
 }
--- a/src/main/java/org/mozilla/gecko/sync/repositories/HashSetStoreTracker.java
+++ b/src/main/java/org/mozilla/gecko/sync/repositories/HashSetStoreTracker.java
@@ -20,6 +20,7 @@ public class HashSetStoreTracker implements StoreTracker {
     guids = new HashSet<String>();
   }
 
+  @Override
   public String toString() {
     return "#<Tracker: " + guids.size() + " guids tracked.>";
   }
--- a/src/main/java/org/mozilla/gecko/sync/repositories/HashSetStoreTracker.java
+++ b/src/main/java/org/mozilla/gecko/sync/repositories/HashSetStoreTracker.java
@@ -25,7 +25,7 @@ public class HashSetStoreTracker implements StoreTracker {
   }
 
   @Override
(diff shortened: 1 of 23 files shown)


ncalexan:
Two general comments that I also brought up on IRC.

I find the term 'tracking' to be really ambiguous.  You could track records for a million reasons, only one of which is not reuploading.  The crucial method is 'isTrackedForExclusion' suggests trackedForExclusion throughout.

I also have an ideological issue with testing things in the callback.  If your callbacks never get called, nothing will happen, but your tests won't fail.  I prefer having the callbacks collect things which are then verified after the action is over.  Then you really know that things happened.  And you can also check that bad things _didn't_ happen.  It's also a lot easier to deduce what you wanted to happen -- rather than reverse engineering when the callbacks are called, you're reading someone assert something about the state of a callback object.

rnewman:
Re testing in callbacks: if the expected callback method isn't called, then it also won't notify the monitor, and the test won't complete. (And the other arms of the delegate all fail noisily by default.)

There's no other part of the code that calls performNotify with the appropriate test helper, so if the test completes it's because it's passed through our delegate methods.

If our tests are passing because we're not waiting for the callback to be called, then we have bigger problems than testing prior to notifying.

I understand your point, and I agree to an extent, but it's often more convenient to do checks where the objects are available for inspection.

src/main/java/org/mozilla/gecko/sync/repositories/HashSetStoreTracker.java
```diff
@@ -0,0 +1,54 @@
...
+  // Used to store GUIDs that were not locally modified but
+  // have been modified by a call to `store`, and thus
+  // should not be returned by a subsequent fetch.
+  private HashSet<String> guids;
+
+  public HashSetStoreTracker() {
+    guids = new HashSet<String>();
+  }
+
+  public String toString() {
```
liuche, line 24: @Override annotation.

src/main/java/org/mozilla/gecko/sync/repositories/RepositorySession.java
```diff
@@ -251,4 +256,91 @@ public void abort() {
...
+  protected Record reconcileRecords(final Record remoteRecord,
+                                    final Record localRecord,
+                                    final long lastRemoteRetrieval,
+                                    final long lastLocalRetrieval) {
+    Log.d(LOG_TAG, "Reconciling remote " + remoteRecord.guid + " against local " + localRecord.guid);
+
+    if (localRecord.equalPayloads(remoteRecord)) {
+      if (remoteRecord.lastModified > localRecord.lastModified) {
+        Log.d(LOG_TAG, "Records are equal. No record application needed.");
+        return null;
```
liuche, line 304: Nit: Both nested if checks return null, can remove this return and just fall through, unless this is for comment clarity.

rnewman, line 304: Yeah, it's for clarity. I want all this logic to be as obvious as possible.

(I already had one bug around fallthrough in WBORepository that my tests caught.)

test/src/org/mozilla/android/sync/test/StoreTrackingTest.java
```diff
@@ -0,0 +1,429 @@
...
+      // Should not happen.
+    }
+  }
+
+  private void doTestNewSessionRetrieveByTime(final WBORepository repository,
+                                              final String expectedGUID) {
+    final SuccessCreationDelegate createDelegate = new SuccessCreationDelegate() {
+      @Override
+      public void onSessionCreated(final RepositorySession session) {
+        System.out.println("Session created.");
```
liuche: Use Log instead? Not sure about when we want to use System.out.println.

rnewman: Yeah, this is test code, so it doesn't really matter, but I should be consistent.

ncalexan:
Looks good to me.

PR #44, opened 2012-01-31
Feedback: Bug 718703 - Pair a Device Crashing

liuche:
Problem: using org.mozilla.gecko package name in the Intent fired off by "Pair a Device" in sync settings, instead of Fennec package. Added fixes in preprocessing, to create correct package reference for both standalone Android Sync and Fennec-bundled Android Sync.

Tested hard-coding package name "org.mozilla.fennec_liuche" when bundling with Fennec, did not crash.

Also made changes to bundling scripts, but wasn't able to actually test on Fennec bundled, because of build problems.

make[6]: **\* No rule to make target `/Users/liuche/moz/code/hg/m-c/mobile/android/base/resources/xml/sync_preferences.xml', needed by`res/layout/sync_account.xml'.  Stop.
make[6]: **\* Waiting for unfinished jobs....
make[5]: **\* [libs] Error 2

rnewman: maybe for when you have time, updating the README for the places that need to be changed in order to add new resources.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ AndroidManifest.xml
 R.java
 src/main/java/org/mozilla/gecko/sync/repositories/android/Authorities.java
 res/xml/sync_syncadapter.xml
+res/xml/sync_preferences.xml
 res/values/strings.xml
 .metadata
 *.pyc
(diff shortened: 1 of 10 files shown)


rnewman:
I cleaned this up and merged. Working on the Makefile stuff now.

PR #47, opened 2012-01-31
Nalexander/bug 722430 make hmac hasher null

ncalexan:
...wrap in CryptoException higher up the call stack.

This one got pretty gnarly, but I think does what bsmith wants.  It propagates NoSuchAlgorithmException and InvalidKeyException higher up the call stack and wraps in CryptoException where appropriate.


rnewman:
Done. I took the opportunity to switch a couple of MPL headers and fix whitespace (from 4 to 2) in a couple of files.

PR #49, opened 2012-02-03
Nalexander/wait helper

ncalexan:
Most important: this tests WaitHelper.  I was confused by what was supposed to happen.

Next important: I made WaitHelper use a length 1 blocking queue.  This makes synchronous callbacks (that don't spawn a thread) work.  This also makes testing and reacting to various conditions a lot easier; hopefully I didn't introduce too many concurrency problems.

Next next important: I am catching more than AssertionError; I am also catching AssertionFailedError.  Something is different between JUnit 3 and JUnit 4, but now I get good stack traces (in Eclipse) without needing to crawl through logcat.

Least important: I added an optional timeout to performWait.  Handy when you are testing and probably a good thing to set globally at some point -- long tests are probably broken, after all, although the emulator does make that dicy.  JUnit 4 will do the timeouts for us.

### DIFF
--- a/test/src/org/mozilla/android/sync/test/WaitHelperTest.java
+++ b/test/src/org/mozilla/android/sync/test/WaitHelperTest.java
@@ -1,0 +1,293 @@
+/* Any copyright is dedicated to the Public Domain.
+   http://creativecommons.org/publicdomain/zero/1.0/ */
+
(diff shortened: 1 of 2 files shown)


rnewman:
Nits fixed and merged.

PR #50, opened 2012-02-03
Bug 723235 - Request history items sorted by sortindex/frecency.

rnewman:
Also covers Bug 720304, with a hard limit of 500 history items per request.

Now to do the same thing (for sanity only) for bookmarks.


rnewman:
Crap. Wrong destination.

PR #51, opened 2012-02-03
Bug 723235 - Request history items sorted by sortindex/frecency.

rnewman:
Also covers:

• Bug 720304, with a hard limit of 500 history items per request.
• Bug 723240, requesting bookmarks sorted by frecency.

and imposes a sanity limit of 5,000 bookmarks per fetch.

(Note that batching is outside the scope of these bugs.)


ncalexan:
+1 from me.

rnewman:
Will address nits in future.

PR #52, opened 2012-02-03
Bug 705767: Validate credentials before completing classic setup (+feedback on http error)

liuche:
Also lifted out account creation and made some setup UI changes. We may want to validate the sync key length, because that is not checked against anything before being added to an account.

HTTP error: after a 401 unauthorized credential validation against the sync server, the next request will always fail with a ClientProtocolError. (throw ProtocolException in http://dxr.mozilla.org/mozilla/mozilla-central/mobile/android/base/httpclientandroidlib/impl/conn/DefaultResponseParser.java.html)
Not consuming the resource after use will make this problem disappear, but also exhausts the thread pool. Some detail about the Java HttpClient that I'm missing?

### DIFF
--- a/res/layout/sync_account.xml
+++ b/res/layout/sync_account.xml
@@ -61,6 +61,7 @@
     android:orientation="horizontal" >
 
     <Button
+      android:id="@+id/accountCancelButton"
       style="@style/SyncButton"
       android:onClick="cancelClickHandler"
       android:text="@string/sync_button_cancel" />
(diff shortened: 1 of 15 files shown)


liuche:
Added error handling for P1 bug 709879 - relies on newly lifted-out AccountCreator, so put it in the same branch.

liuche:
Updated and rebased in branch liuche/fix-setup.

PR #53, opened 2012-02-03
Nalexander/bug 722524 j pake crypto check zkp off by one

ncalexan:


src/main/java/org/mozilla/gecko/sync/jpake/Gx3OrGx4IsZeroOrOneException.java
```diff
@@ -37,6 +37,6 @@
 
 package org.mozilla.gecko.sync.jpake;
 
-public class Gx4IsOneException extends Exception {
-  private static final long serialVersionUID = 815366241052296473L;
+public class Gx3OrGx4IsZeroOrOneException extends IncorrectZkpException {
```
liuche: Why extend IncorrectZkpException? The isZeroOrOne check is unrelated to ZKP, it's an extra check that the J-PAKE protocol requires.

rnewman:
Looks good to me.

PR #59, opened 2012-02-07
Bug 722541 - always reseed with 8 random bytes.

ncalexan:
I confused the number of bytes to seed from with the seed itself.  This prevents massive memory allocation as reported by liuche.


ncalexan:
Relevant: http://www.cigital.com/justice-league-blog/2009/08/14/proper-use-of-javas-securerandom/

rnewman:
Folded into the original patch for when this lands in m-c.

PR #63, opened 2012-02-08
Cleaning

ncalexan:
No broken windows, at least in our packages.


rnewman:
I merged this. Git should happily realize that when you rebase your crypto branch.

PR #64, opened 2012-02-08
Remove Cryptographer.

ncalexan:
Further refactoring.  Sorry that this builds on other branches...


rnewman:
I stopped reviewing after the first four commits. Please make sure that you're not reducing test coverage as you simplify; if in doubt, I'd much rather have tests that you think might overlap than lose tests that don't.

ncalexan:
Everything here has been merged through nalexander/remove-(Sync)?Cryptographer-prime.

PR #66, opened 2012-02-13
Bug 705767: Validate credentials before completing classic setup + fixes

liuche:
Added check for user existence with 1.0/user; use server URL if info/collections then returns 404.

HTTP request to server is set to be "kept alive indefinitely" and isn't consumed on consumeEntity - leads to next HTTP request connection getting shut down. Not sure what to do about that.

### DIFF
--- a/res/layout/sync_account.xml
+++ b/res/layout/sync_account.xml
@@ -61,6 +61,7 @@
     android:orientation="horizontal" >
 
     <Button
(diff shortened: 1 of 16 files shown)


liuche:
Reordering of 1213a08 was caused by me reverting an earlier version of a file incorrectly...

liuche:
Updated pull request: https://github.com/mozilla-services/android-sync/pull/74

PR #68, opened 2012-02-14
Bug 717137: enable landscape view in J-PAKE.

liuche:
Enable landscape view, at the cost of not automatically advancing editable text boxes during PIN entry in Pair a Device.


rnewman:
That "at the cost of" sounds pretty sinister. What's the way out?

PR #70, opened 2012-02-14
Bug 726054 - Desktop parity: track last sync time from final upload, not final fetch

rnewman:
This sequence of commits threads a store finish time through the delegate callbacks, mirroring fetch, and isolates logic for deciding on the final persisted timestamp in a single location in Synchronizer.

It also includes a trivial extension of an existing test.


ncalexan:
Fine by me.

rnewman:
Merged with review comments.

PR #72, opened 2012-02-15
Bug 722007 - update row in place in AndroidBrowserHistoryDataExtender.

ncalexan:
And factor out AndroidBrowserHistoryDataExtenderTest.

The ticket asks not to blow away the existing visit data, but I don't know how to do that without changing the history extension database schema: since visits are stored as a JSON string, there's no 'update only part'.  I doubt we want to change the database schema right now, so I'm pushing this small cleaning/preparation for new data extenders patch and suggesting we close/de-prioritize 722007.

Or I've completely misunderstood :)

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserHistoryDataExtender.java
+++ b/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserHistoryDataExtender.java
@@ -139,17 +139,6 @@ public class AndroidBrowserHistoryDataExtender extends SQLiteOpenHelper {
   public void store(String guid, JSONArray visits) {
     SQLiteDatabase db = this.getCachedWritableDatabase();
 
-    boolean rowExists = false;
-    try {
-      Cursor cur = fetch(guid);
-      if (cur.getCount() >= 1) {
-        rowExists = true;
-      }
-      cur.close();
-    } catch (NullCursorException e) {
-      // Do nothing.
-    }
-
     ContentValues cv = new ContentValues();
     cv.put(COL_GUID, guid);
     if (visits == null) {
@@ -158,15 +147,14 @@ public class AndroidBrowserHistoryDataExtender extends SQLiteOpenHelper {
       cv.put(COL_VISITS, visits.toJSONString());
     }
 
-    if (!rowExists) {
+    String where = COL_GUID + " = ?";
+    String[] args = new String[] { guid };
+    int rowsUpdated = db.update(TBL_HISTORY_EXT, cv, where, args);
+    if (rowsUpdated >= 1) {
+      Logger.debug(LOG_TAG, "Replaced history extension record for row with GUID " + guid);
+    } else {
       long rowId = db.insert(TBL_HISTORY_EXT, null, cv);
       Logger.debug(LOG_TAG, "Inserted history extension record into row: " + rowId);
-    } else {
-      String where = COL_GUID + " = ?";
-      String[] args = new String[] { guid };
-
-      db.update(TBL_HISTORY_EXT, cv, where, args);
-      Logger.debug(LOG_TAG, "Replaced history extension record for row with GUID " + guid);
     }
   }
 
--- a/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserHistoryDataExtender.java
+++ b/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserHistoryDataExtender.java
@@ -61,8 +61,11 @@ public class AndroidBrowserHistoryDataExtender extends SQLiteOpenHelper {
   public static final String COL_GUID = "guid";
   public static final String COL_VISITS = "visits";
 
+  private final RepoUtils.QueryHelper queryHelper;
+
   public AndroidBrowserHistoryDataExtender(Context context) {
     super(context, DB_NAME, null, SCHEMA_VERSION);
+    this.queryHelper = new RepoUtils.QueryHelper(context, null, LOG_TAG);
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


emtwo:
Here's the original commit with a few comments: https://github.com/mozilla-services/android-sync/commit/b22ef

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserHistoryDataExtender.java
```diff
@@ -46,12 +47,11 @@
 import android.database.Cursor;
 import android.database.sqlite.SQLiteDatabase;
 import android.database.sqlite.SQLiteOpenHelper;
-import android.util.Log;
 
 public class AndroidBrowserHistoryDataExtender extends SQLiteOpenHelper {
 
-  public static final String TAG = "AndroidBrowserHistoryDataExtender";
-  
+  public static final String LOG_TAG = "A.BrowHist.DataExtender";
```
rnewman: Let's go with <tt>SyncHistoryVisits</tt>.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserHistoryDataExtender.java
```diff
@@ -123,15 +126,29 @@ public void wipe() {
...
+   * Store visit data.
+   *
+   * If a row with GUID `guid` does not exist, insert a new row.
+   * If a row with GUID `guid` does exist, replace the visits column.
+   *
+   * @param guid The GUID to store to.
+   * @param visits New visits data.
+   */
+  public void store(String guid, JSONArray visits) {
     SQLiteDatabase db = this.getCachedReadableDatabase();
```
rnewman: _blink_

I'm pretty sure that should be a WritableDatabase…

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserHistoryDataExtender.java
```diff
@@ -139,34 +156,80 @@ public long store(String guid, JSONArray visits) {
     } else {
       cv.put(COL_VISITS, visits.toJSONString());
     }
-    long rowId = db.insert(TBL_HISTORY_EXT, null, cv);
-    Log.i(TAG, "Inserted history extension record into row: " + rowId);
-    return rowId;
+
+    if (!rowExists) {
```
rnewman: No need for this. Do something like this:

<pre>
int updated = db.update(TBL_HISTORY_EXT, cv, where, args);
if (updated > 0) {
  // Log
  return;
}
db.insert(TBL_HISTORY_EXT, null, cv);
</pre>


The update _is_ a query, followed by an in-place update on the cursor within the same transaction. Might as well save the round-trip and such.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserHistoryDataExtender.java
```diff
@@ -139,34 +156,80 @@ public long store(String guid, JSONArray visits) {
...
-                          new String[] { COL_GUID, COL_VISITS },
-                          where, args,
-                          null, null, null);
-    RepoUtils.queryTimeLogger("AndroidBrowserHistoryDataExtender.fetch(guid)", queryStart, System.currentTimeMillis());
-    if (cur == null) {
-      Log.e(TAG, "Got a null cursor while doing fetch for guid " + guid + " on history extension table");
-      throw new NullCursorException(null);
-    }
+    Cursor cur = queryHelper.safeQuery(db, ".fetch", TBL_HISTORY_EXT,
+        new String[] { COL_GUID, COL_VISITS },
```
rnewman: Break this out into a const. No need to allocate a new array.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserHistoryDataExtender.java
```diff
@@ -139,34 +156,80 @@ public long store(String guid, JSONArray visits) {
...
+  /**
+   * Fetch all rows.
+   *
+   * @return A Cursor.
+   * @throws NullCursorException
+   */
+  public Cursor fetchAll() throws NullCursorException {
+    SQLiteDatabase db = this.getCachedReadableDatabase();
+    Cursor cur = db.query(TBL_HISTORY_EXT,
+        new String[] { COL_GUID, COL_VISITS },
```
rnewman: Again, const.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserHistoryDataExtender.java
```diff
@@ -139,34 +156,80 @@ public long store(String guid, JSONArray visits) {
...
+        null, null, null, null, null);
+    if (cur == null) {
+      throw new NullCursorException(null);
```
rnewman: Can we just use <tt>safeQuery</tt> here?

test/src/org/mozilla/gecko/sync/repositories/android/test/AndroidBrowserHistoryDataExtenderTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+import android.content.Context;
+import android.database.Cursor;
+import android.test.ActivityInstrumentationTestCase2;
+import android.util.Log;
+
+public class AndroidBrowserHistoryDataExtenderTest extends ActivityInstrumentationTestCase2<StubActivity> {
+
+  protected AndroidBrowserHistoryDataExtender extender;
+  protected static final String LOG_TAG = "AndroidBrowserHistoryDataExtenderTest";
```
rnewman: Needs to be shorter than 23 characters.

PR #75, opened 2012-02-17
Bug 728374 - update inconsistent licenses.

ncalexan:
- First commit: a tool to check_headers.sh backed by tools/check_headers.py.

Try `check_headers.sh -d -v` to get a verbose dry-run and see what's up.
- Second commit: result of running tool.  A few comments are stripped, but I think this is actually a good thing :)  (Search for '// ' to see them.)


rnewman:
Bug-named and merged the first commit.

PR #76, opened 2012-02-19
Bug 728636 - Conversion of email address to account should lowercase

rnewman:


ncalexan:
Changes as written look fine.  I didn't try to verify that you hit all places this lowercase change might be required.

PR #79, opened 2012-02-21
Bug 729273 - Don't impose modified times on stored records.

rnewman:


rnewman:
Merged with comment clarified. Thanks!

PR #81, opened 2012-02-22
Bug 729410 - Implement basic automatic backoff unless a sync is forced.

rnewman:
Android appears to sync every minute or so. That's baaaaad. This patch ensures a minimum backoff of 5 minutes between syncs, unless you force it through the Settings UI.

Fennec will be able to get around this.


rnewman:
Landed with bookmarks.

PR #84, opened 2012-02-23
Bug 705767: Validate credentials before completing classic setup + rebase

liuche:
One last issue I haven't been able to fix - a failed auth attempt 401 causes the HttpRequest connection to be "kept alive indefinitely." The subsequent auth request will always hang.

Logcat:

02-23 14:32:02.906: I/BaseResource(4246): Response: HTTP/1.1 401 Unauthorized
02-23 14:32:12.286: I/BaseResource(4246): HTTP GET https://auth.services.mozilla.com/user/1.0/7i77ycjdvltg7xfbsooj3acmdh6mbzbz
02-23 14:32:12.286: D/class ch.boye.httpclientandroidlib.impl.client.DefaultHttpClient(4246): Stale connection check
02-23 14:32:12.286: D/class ch.boye.httpclientandroidlib.client.protocol.RequestAuthCache(4246): Auth cache not set in the context
02-23 14:32:12.528: I/BaseResource(4246): Response: HTTP/1.1 200 OK
02-23 14:32:12.536: I/BaseResource(4246): HTTP GET https://auth.services.mozilla.com/user/1.0/7i77ycjdvltg7xfbsooj3acmdh6mbzbz/node/weave
02-23 14:32:12.546: D/class ch.boye.httpclientandroidlib.impl.client.DefaultHttpClient(4246): Stale connection check
02-23 14:32:12.556: D/class ch.boye.httpclientandroidlib.client.protocol.RequestAuthCache(4246): Auth cache not set in the context
02-23 14:32:12.676: I/BaseResource(4246): Response: HTTP/1.1 200 OK
02-23 14:32:12.686: I/BaseResource(4246): HTTP GET https://scl2-sync395.services.mozilla.com/1.1/7i77ycjdvltg7xfbsooj3acmdh6mbzbz/info/collections/
02-23 14:32:12.706: D/class ch.boye.httpclientandroidlib.impl.client.DefaultHttpClient(4246): Stale connection check
02-23 14:32:12.716: D/class ch.boye.httpclientandroidlib.client.protocol.RequestAuthCache(4246): Auth cache not set in the context
02-23 14:32:12.716: D/class ch.boye.httpclientandroidlib.impl.client.DefaultHttpClient(4246): Attempt 1 to execute request
02-23 14:32:12.726: D/class ch.boye.httpclientandroidlib.impl.conn.DefaultClientConnection(4246): Connection closed
02-23 14:32:12.726: D/class ch.boye.httpclientandroidlib.impl.conn.DefaultClientConnection(4246): Connection shut down

(followed by hang)

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/setup/auth/AuthenticateAccountStage.java
+++ b/src/main/java/org/mozilla/gecko/sync/setup/auth/AuthenticateAccountStage.java
@@ -68,6 +68,7 @@ public class AuthenticateAccountStage implements AuthenticatorStage {
     String authHash = Base64.encodeToString((aa.usernameHash + ":" + aa.password).getBytes(), Base64.DEFAULT);
     String authRequestUrl = aa.authServer + Constants.AUTH_SERVER_VERSION + aa.usernameHash + "/" + Constants.AUTH_SERVER_SUFFIX;
     authenticateAccount(callbackDelegate, authRequestUrl, authHash);
+
   }
 
   private void authenticateAccount(final AuthenticateAccountStageDelegate callbackDelegate, final String authRequestUrl, final String authHeader) throws URISyntaxException {
--- a/src/main/java/org/mozilla/gecko/sync/setup/auth/AuthenticateAccountStage.java
+++ b/src/main/java/org/mozilla/gecko/sync/setup/auth/AuthenticateAccountStage.java
@@ -23,6 +23,7 @@ import ch.boye.httpclientandroidlib.message.BasicHeader;
 
 public class AuthenticateAccountStage implements AuthenticatorStage {
   private final String LOG_TAG = "AuthenticateAccountStage";
+  private HttpRequestBase httpRequest = null;
 
   public interface AuthenticateAccountStageDelegate {
     public void handleSuccess(boolean isSuccess);
@@ -44,14 +45,21 @@ public class AuthenticateAccountStage implements AuthenticatorStage {
       public void handleFailure(HttpResponse response) {
         Logger.debug(LOG_TAG, "handleFailure");
         aa.abort(response.toString(), new Exception(response.getStatusLine().getStatusCode() + " error."));
+        if (response.getEntity() == null) {
+          // No cleanup necessary.
+          return;
+        }
         try {
           BufferedReader reader = new BufferedReader(new InputStreamReader(response.getEntity().getContent(), "UTF-8"));
           Logger.warn(LOG_TAG, "content: " + reader.readLine());
           SyncResourceDelegate.consumeReader(reader);
-          reader.close();
-          SyncResourceDelegate.consumeEntity(response.getEntity());
         } catch (IllegalStateException e) {
           Logger.debug(LOG_TAG, "Error reading content.", e);
+        } catch (RuntimeException e) {
+          Logger.debug(LOG_TAG, "Unexpected exception.", e);
+          if (httpRequest != null) {
+            httpRequest.abort();
+          }
         } catch (IOException e) {
           Logger.debug(LOG_TAG, "Error reading content.", e);
         }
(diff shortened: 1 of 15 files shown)


liuche:
Overriding connectionTimeout and socketTimeout to return a timeout value (e.g., 5*1000) does not seem to work, and no timeout exceptions are thrown.

emtwo:
I haven't thoroughly looked at the code yet, but I had some hanging problems too and they were caused by not consuming the http body or entity. Make sure this happens in the 401 case?

liuche:
Thanks for the suggestion; there is already a SyncResourceDelegate.consumeEntity on the HTTP resource after handling the 401 case.

src/main/java/org/mozilla/gecko/sync/setup/auth/AuthenticateAccountStage.java
```diff
@@ -0,0 +1,136 @@
...
+        aa.isSuccess = isSuccess;
+        aa.runNextStage();
+      }
+
+      @Override
+      public void handleFailure(HttpResponse response) {
+        Logger.debug(LOG_TAG, "handleFailure");
+        aa.abort(response.toString(), new Exception(response.getStatusLine().getStatusCode() + " error."));
+        try {
+          BufferedReader reader = new BufferedReader(new InputStreamReader(response.getEntity().getContent(), "UTF-8"));
```
rnewman, line 57: Firstly, you want to check for null.

```
HttpEntity entity = response.getEntity();
if (entity == null) {
  // No cleanup needed.
  return;
}
```

Secondly, you might be getting a `RuntimeException` here. Add a clause:

```
} catch (RuntimeException e) {
  Logger.warn(LOG_TAG, "Unexpected exception.", e);
  // TODO: get a handle to the HttpGet itself, and call .abort() on it.
}
```

src/main/java/org/mozilla/gecko/sync/setup/auth/AuthenticateAccountStage.java
```diff
@@ -0,0 +1,136 @@
...
+          Logger.warn(LOG_TAG, "content: " + reader.readLine());
+          SyncResourceDelegate.consumeReader(reader);
+          reader.close();
+          SyncResourceDelegate.consumeEntity(response.getEntity());
```
rnewman: Thirdly, this doesn't make any sense at all. You're already consuming the reader, which will consume the underlying `InputStream`. I wouldn't be surprised if this were hanging or throwing a `RuntimeException`.

Do one of three things:
1. Call `abort` on the `HttpGet`.
2. If you have a `Reader` over the stream, consume that.
3. If you don't, consume the entity.

Don't do more than one :)

src/main/java/org/mozilla/gecko/sync/setup/auth/AuthenticatorStage.java
```diff
@@ -0,0 +1,8 @@
+package org.mozilla.gecko.sync.setup.auth;
```
rnewman, line 5: MPL on all these files.

liuche:
Added fixes, but still encounter hanging. The get request is made, but the delegate never handles the response in the hanging cases. Is the request not being made/sent correctly?

rnewman:
Do you have an automated test for this? You should be able to write a unit test for `AuthenticateAccountStage` using the HTTP server that we use. And then we can make it pass.

Also, you need to rebase — note in particular the non-trivial changes in `AccountActivity.java`:

```
    // Set components to sync (default: all).
    ContentResolver.setMasterSyncAutomatically(true);

    String authority = BrowserContract.AUTHORITY;
    Log.d(LOG_TAG, "Setting authority " + authority + " to sync automatically.");
    ContentResolver.setSyncAutomatically(account, authority, true);
    ContentResolver.setIsSyncable(account, authority, 1);
```

rnewman:
Chenxia: could you rebase this (see my last comment, too) and see if you still have problems after Nick's recent stream fixes?

liuche:
Rebased into liuche/merge-setup-rebase.

Tested for hanging and it is still present. Will write unit test for the try-retry-credentials case.

rnewman:
> Tested for hanging and it is still present. Will write unit test for the try-retry-credentials case.

Let me know when you've got something you want me to look at.

rnewman:
… and open a pull request for it when you do! Just noticed this one was closed.

PR #86, opened 2012-02-24
Bug 730189 - Fix internet connection check.

liuche:
Check all networks for connectivity, instead of just wifi and mobile data.

My phone does not have bluetooth tethering, so have not tested with non-mobile/wifi networks.

Alternatively, can check ActiveNetwork instead of iterating through all networks, but this should catch all cases.


rnewman:
Merged.

PR #87, opened 2012-02-25
Bug 730557 - Use Fennec BrowserContract directly

rnewman:
Based on work Nick started in his forms engine branch.

With the password tests disabled, this passes all 199 tests. \o/

It also builds unmodified in m-i; we're set up to use the same BrowserContract in each location, no package munging necessary.

I also took the time to make a couple of fixes to the deployment script.


ncalexan:
Can we get rid of Authorities entirely, or fold it into BrowserContractHelpers?

rnewman:
We can get rid of the current use of Authorities, but we do need a preprocessed file for our own use (e.g., for client name generation), so it can't go entirely.

ncalexan:
No sense making a change then.  Thanks!

rnewman:
Landed on inbound, including Authorities => GlobalConstants.

PR #88, opened 2012-02-26
Bug 730637 - Base64 support for HttpClient in Android 2.1.

rnewman:
This adds a shell script to undo the transformation that's built-in to the HttpClient release we use. Instead, we use our own Apache Commons distribution. Works for me on ICS/Transformer; I've asked nhirata to test in

https://bugzilla.mozilla.org/show_bug.cgi?id=730637


rnewman:
Merged and landed on inbound.

PR #91, opened 2012-02-28
Bug 723230 - secreview meta bug.

ncalexan:
From the "it was there" files: try to move on a few of these outstanding P1's.  Also commented on open tickets blocking 723230.


rnewman:
It makes `counter` _safer_, because it's only evaluated once. It doesn't make it `safe`, because accesses to `long`s in Java aren't atomic.

For a logging operation, I'm happy with "fairly safe", which is why I didn't synchronize this counter when I wrote the original code. The overhead of synchronization or an `AtomicLong` is unnecessary. `volatile` would address the issue.

rnewman:
r=me with `volatile` for the counter and `private` for `enable…`. Merge when done.

rnewman:
Actually, hold off on the merge, please. I want to get Marina's stuff in in one go, so I'll merge this afterwards. Thanks!

ncalexan:
I rebased to develop, borking the original commit comments :(  I hate that.

rnewman:
Good good. Could you flatten the review comments into the main commits? Makes it easier for me to land on m-i.

PR #93, opened 2012-03-02
Bug 732114 - move BookmarkHelper stuff out of AndroidBrowserRepositoryTest.

ncalexan:
I tried to keep the changes fairly small, but given your comment on the bug, perhaps you intended AndroidBrowserRepositoryTest to only work for Bookmarks (and History?).  Frankly, I'm surprised it does work for History; it fails because of the SPECIAL_GUIDS_MAP for my FormHistory tests.  Maybe the special guids are always present for History repos, and we always ignore the history for the special guids?


ncalexan:
Uh, github is saying I added these commits 17 days ago.  Does it say that for anyone else?  Fascinating, my git repo says that too.  A problem with my Emacs/git interface, perhaps?  Weird.

rnewman:
Looking good with minor nits, but I'd like to see the `Simple*` delegates adjusted to be independent before this lands. I'd like to use them in my tabs tests, which aren't related to `AndroidBrowser*`.

PR #90, opened 2012-02-28
Bug729248 - Smarter upload of our clients record.

emtwo:
Needs a couple of more tests, but a quick review would be nice.

PR #95, opened 2012-03-02
Bug729248 - Smarter upload of our clients record (rebased).

emtwo:


emtwo:
Only the commit titled "Bug 729248 - Smarter upload of our clients record." Is relevant. I'm not sure why github wouldn't let me only pull-request that commit even though it's fully rebased.

The commit is rebased on top of marina/rebased-clients-stage and addresses comments in https://github.com/mozilla-services/android-sync/pull/90.
It's rough but is simply a proposal for what to do when upload fails.

Next things I want to do/consider:
- split into 2 stages
- What to do 2nd time commands are downloaded (probably will handle in commands bug)
- What to do if it fails more than the maximum number of failures? Maybe we shouldn't have a max fail? Maybe not a problem if we re-process the same commands since we'll handle this in the previous point?

emtwo:
Moved to https://github.com/mozilla-services/android-sync/pull/99

PR #94, opened 2012-03-02
Bug 732114 rebased.

ncalexan:
See https://github.com/mozilla-services/android-sync/pull/93.

I rebase-`fixup`-ed reviewer comments into the appropriate patches.

I made the Simple\* delegates inherit from `DefaultDelegate`, added a constructor specifying which `WaitHelper` to use (by default, `WaitHelper.getTestWaiter()`), and generally seperated these delegates from AndroidBrowser*.

PR #97, opened 2012-03-04
Bug 732768 - Eliminate shared state and concurrency problems in tests

rnewman:
See https://bugzilla.mozilla.org/show_bug.cgi?id=732768.

I was having some annoyances with tests that led me through a rats nest of problems.

I fixed many of them.

Improvements:
- We now shut down our worker queues in finish. This exposed a whole raft of bugs; we can't, for example, delegate the failure callbacks for when the queues are shut down!
- We now throw exceptions if our state is incorrect to handle a method. That means you get a synchronous exception if you try to finish twice, or begin twice, or begin after finish.
- There was no synchronization around a single shared session slot. Enormous possibility for error. Now sessions are passed around, and there is no global RepositorySession in tests; the only remaining global test helper is the wait helper itself.
- Callbacks are now deferred correctly in more places.
- Some methods that should notify — e.g., finish! — did not. They're now Runnables.
- I lifted the shared middleware stuff out of Crypto5MiddlewareRepository.
- Session state wasn't safe. Now it is, and operations such as "run this runnable if we're active and our queue isn't shut down" are now atomic.
- Items could be left in the WaitHelper queue between tests. I added a sanity method to assert if that's the case. It caught bugs.

Sorry that this is going to be an ugly pull request, but this was all very tangled, and I eventually gave up trying to produce incremental commits.


rnewman:
I'll eventually make the other methods (besides `fetch`) check for activity, too. It's annoying to have to sprinkle `catch` blocks everywhere… but on the other hand, wherever there isn't a catch block there's a chunk of code that will allow the thread to die if an exception occurs.

ncalexan:
A quick look up to the patch bomb (5931b56) looks fine.  I'll give everything a closer reading Monday, and try to rebase/incorporate/discard https://github.com/mozilla-services/android-sync/pull/94.  Would have been much better if you had worked on top of 94 :(

rnewman:
Yeah, I got half-way through, went to look for your pull request, and it didn't appear in the list, so I rolled on regardless. Overly enthusiastic caching somewhere. 

I'm happy to take care of rebasing.

rnewman:
Rebase done:

https://github.com/mozilla-services/android-sync/pull/98

PR #102, opened 2012-03-09
Bug 734521 - consume response audit

ncalexan:


indygreg:
I would argue that running tests in a non-deterministic order is better. Otherwise, you may inadvertently rely on state between tests.

ncalexan:
Hi Greg, I think I agree.

In this case random order actually helped hide the problem -- it was an interaction between lots of tests being run in a certain order by maven that failed (including alphabetical) and almost any other order succeeded.  Happy to fixup and have this random.

indygreg:
Wow, I think you hit the exception.

ncalexan:
Running tests in random order landed as part of Bug 709329; putting consumeEntity helpers added as separate part.

rnewman:
Merged.

PR #108, opened 2012-03-15
Tool for dumping Fennec databases.

ncalexan:
Very rough, but WFM on Transformer Prime.  May cause blindness, impotence, or blindness and impotence.


rnewman:
If this doesn't work, I don't even care. Awesome.

PR #109, opened 2012-03-15
Bug 735996 - HTTP layer improvements

rnewman:
Spotted a few of these as I was working on the clients stuff.
- Set an AuthCache to avoid annoying log messages.
- Don't check for stale connections on every request (10-100ms overhead); have a reaper thread do the work.
- Adjust connection pool limits.
- Terminate connection pool at the end of a sync.
- Close streams, don't completely consume them.
- Misc.


ncalexan:
+1 from me.

PR #111, opened 2012-03-15
Bug 715792 - Process client record commands.

emtwo:
This is pretty rough & still doesn't have tests but I wanted some input. @rnewman I used the "register your command" idea we talked about a couple of days ago, but not sure if I fully captured it correctly.

This approach limits a single command per class. So `GlobalSession` cannot have more than one command, for example, unless we have a nested class for each method.

Thoughts, ideas, suggestions?

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/CommandProcessor.java
+++ b/src/main/java/org/mozilla/gecko/sync/CommandProcessor.java
@@ -1,0 +1,83 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.gecko.sync;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.json.simple.JSONArray;
+
+public class CommandProcessor {
+  private static final String LOG_TAG = "Command";
+  protected ConcurrentHashMap<String, CommandRunner> commands = new ConcurrentHashMap<String, CommandRunner>();
+
+  private final static CommandProcessor processor = new CommandProcessor();
+
+  public static CommandProcessor getProcessor() {
+    return processor;
+  }
+
+  public static class Command {
+    public final String commandType;
+    public final List<String> args;
+
+    public Command(String commandType, List<String> args) {
+      this.commandType = commandType;
+      this.args = args;
+    }
+  }
+
+  public void registerCommand(String commandType, CommandRunner command) {
+    commands.put(commandType, command);
+  }
+
+  public void processCommand(ExtendedJSONObject unparsedCommand) {
+    Command command = parseCommand(unparsedCommand);
+    if (command == null) {
+      Logger.debug(LOG_TAG, "Invalid command: " + unparsedCommand + " will not be processed.");
+      return;
+    }
+
+    CommandRunner executableCommand = commands.get(command.commandType);
+    if (executableCommand == null) {
+      Logger.debug(LOG_TAG, "Command \"" + command.commandType + "\" not registered and will not be processed.");
+      return;
+    }
+
+    executableCommand.executeCommand(command.args);
+  }
+
+  /**
+   * Parse a JSON command into a ParsedCommand object for easier handling.
+   *
+   * @param unparsedCommand - command as ExtendedJSONObject
+   * @return - null if command is invalid, else return ParsedCommand with
+   *           no null attributes.
+   */
+  protected Command parseCommand(ExtendedJSONObject unparsedCommand) {
+    String type = (String) unparsedCommand.get("command");
+    if (type == null) {
+      return null;
+    }
+
+    try {
+      JSONArray unparsedArgs = unparsedCommand.getArray("args");
+      if (unparsedArgs == null) {
+        return null;
+      }
+      ArrayList<String> args = new ArrayList<String>(unparsedArgs.size());
+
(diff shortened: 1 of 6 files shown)


src/main/java/org/mozilla/gecko/sync/Command.java
```diff
@@ -0,0 +1,11 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.gecko.sync;
+
+import java.util.ArrayList;
+
+public interface Command {
+  public void executeCommand(ArrayList<String> args);
```
rnewman: Just make this `List<String>`.

src/main/java/org/mozilla/gecko/sync/CommandProcessor.java
```diff
@@ -0,0 +1,84 @@
...
+package org.mozilla.gecko.sync;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+
+import org.json.simple.JSONArray;
+
+public class CommandProcessor {
+  private static final String LOG_TAG = "Command";
+  protected HashMap<String, Command> commands = new HashMap<String, Command>();
```
rnewman: `ConcurrentHashMap`.

src/main/java/org/mozilla/gecko/sync/CommandProcessor.java
```diff
@@ -0,0 +1,84 @@
...
+    private final String commandType;
+    private final ArrayList<String> args;
+
+    public ParsedCommand(String commandType, ArrayList<String> args) {
+      this.commandType = commandType;
+      this.args = args;
+    }
+  }
+
+  private static CommandProcessor processor;
```
rnewman: `private final static CommandProcessor processor = new CommandProcessor();`. Then `getProcessor` becomes real easy :D

src/main/java/org/mozilla/gecko/sync/CommandProcessor.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  public void processCommand(ExtendedJSONObject unparsedCommand) {
+    ParsedCommand command = parseCommand(unparsedCommand);
+    if (command == null) {
+      Logger.debug(LOG_TAG, "Invalid command: " + unparsedCommand + " will not be processed.");
+      return;
+    }
+
+    Command executableCommand = commands.get(command.commandType);
```
rnewman: I think a `Command` is the thing that gets run, and a `CommandRunner` is the thing that runs it. I _think_ you're using `ParsedCommand` and `Command` for these two concepts. Can you pick better names?

src/main/java/org/mozilla/gecko/sync/CommandProcessor.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    ArrayList<String> args = new ArrayList<String>();
+    try {
+      JSONArray unparsedArgs = unparsedCommand.getArray("args");
+      if (unparsedArgs == null) {
+        return null;
+      }
+      for (int i = 0; i < unparsedArgs.size(); i++) {
+        args.add((String) unparsedArgs.get(i));
```
rnewman: `unparsedArgs.getString()`

src/main/java/org/mozilla/gecko/sync/CommandProcessor.java
```diff
@@ -0,0 +1,84 @@
...
+   * @return - null if command is invalid, else return ParsedCommand with
+   *           no null attributes.
+   */
+  protected ParsedCommand parseCommand(ExtendedJSONObject unparsedCommand) {
+    String type = (String) unparsedCommand.get("command");
+    if (type == null) {
+      return null;
+    }
+
+    ArrayList<String> args = new ArrayList<String>();
```
rnewman: You can hint the `ArrayList` constructor with `unparsedArgs.size()`.

rnewman:
> This approach limits a single command per class. 

There are three ways to resolve this.
- Have `executeCommand` include the whole command as input. Then you can dispatch accordingly.
- Implement specialized command processors. `ResetEngineCommand`, for example. You can pass an instance of whatever class it needs to poke at: `registerCommand(new ResetEngineCommand(session))`.
- Use anonymous classes to wrap methods: `registerCommand(new Command() { @Override public void executeCommand…`. That way you can access members of the registering instance.

src/main/java/org/mozilla/gecko/sync/CommandProcessor.java
emtwo: You're right, I somehow always manage to pick the most confusing variable names! I'll use `Command` and `CommandRunner` as you suggested

emtwo:
- Addressed comments
- Added tests

src/main/java/org/mozilla/gecko/sync/syncadapter/SyncAdapter.java
```diff
@@ -62,6 +66,15 @@ public SyncAdapter(Context context, boolean autoInitialize) {
...
     mAccountManager = AccountManager.get(context);
+
+    // Register the displayURI command here so our SyncService
+    // can receive notifications to open a URI.
+    CommandProcessor.getProcessor().registerCommand("displayURI", new CommandRunner() {
+      @Override
+      public void executeCommand(List<String> args) {
+        displayURI(args.get(0), args.get(1));
+      }
+    });
```
emtwo, line 77: Might want to move this out into an Activity which has access to the SyncService instead

emtwo:
Rebased.

PR #112, opened 2012-03-16
Bug 736352 - Aborting… doesn't.

rnewman:
With these changes we now terminate outstanding HTTP connections, don't continue processing line-by-line content, don't continue attempting to store, and so on, as soon as we abort the sync.

Before:

https://bugzilla.mozilla.org/attachment.cgi?id=606435

After:

https://bugzilla.mozilla.org/attachment.cgi?id=606482

The only reason I observed this is because we now terminate the HTTP channels when the session is aborted after a single store failure! Previously we were aborting the sync, but it would continue and finish!

Pretty important.


emtwo:
I wouldn't say I fully understand every detail of what's going on here, but the changes do make sense to me. Based on the log, this looks like we're actually aborting when we need to now, unlike before.

+1 from me pending we're all in agreement.

rnewman:
Rebased.

PR #115, opened 2012-03-19
Bug 736972 - java.lang.OutOfMemoryError: at java.io.BufferedReader.<init>(BufferedReader.java)

rnewman:
We were allocating a 2MB buffer when 16KB would suffice.


rnewman:
Yeah, I cheated and put another bug in prior to Part 0.

ncalexan:
Is the 16k chars in the spec?  If so, link?  Otherwise, r+.

rnewman:
Nope, this is just telling Java how big a buffer to use in the BufferedReader. Bigger requires fewer buffer fills per read, smaller saves memory.

PR #117, opened 2012-03-20
Bug 709310 - handle backoff headers on success

ncalexan:


src/main/java/org/mozilla/gecko/sync/net/HttpResponseObserver.java
```diff
@@ -0,0 +1,13 @@
+package org.mozilla.gecko.sync.net;
```
rnewman, line 5: MPL.

src/main/java/org/mozilla/gecko/sync/stage/ServerSyncStage.java
```diff
@@ -158,7 +188,85 @@ public void onSynchronizeFailed(Synchronizer synchronizer,
...
+      requestBackoff(largestBackoffObserved);
+    }
+
     session.abort(null, "Synchronization was aborted.");
   }
+
+  /**
+   * The longest backoff observed to date; -1 means no backoff observed.
+   */
+  protected long largestBackoffObserved = -1;
```
rnewman, line 205: `AtomicLong`, and lose the synchronization below, or add synchronization to `onSynchronizeFailed`/`onSynchronizeAborted`.

rnewman:
As discussed: lift the aborting etc. logic into GlobalSession. Just have it watch HTTP requests, track the max backoff, and decide whether to run the next stage in `advance`.

PR #118, opened 2012-03-20
Bug 731024 - Unsupported record types causes Android Sync to reorder folders containing livemarks/queries/separators…

rnewman:
Still in progress, but tests pass.


ncalexan:
r+ on part 0-3.

rnewman:
Tested: this correctly preserves separator and livemark positions on desktop, even through deletions.

Note part 5: if a remote record has changed, and it's been deleted locally, we use the usual replace/update mechanism to apply it. Part 4 added logic that would check type codes here, which of course is impossible for a mostly blank deleted record.

rnewman:
Oh, and this all fixes bugs in parts 0-3, of course :D

ncalexan:
r+ from me.

PR #119, opened 2012-03-21
Bug 732147 - Send tab to device: receiving pushed tabs.

emtwo:
This works but likely need UX polishing


emtwo:
Notes from 1:1 with rnewman:
- Flash a light along with a received notification (less annoying than vibrating would be)
- See if it's possible within android settings to turn off the flashing light upon notification
- Sync only clients on startup to receive tabs that may have been sent (rnewman: you said there was a related bug blocking 726055, but I'm not sure which one you're talking about)
- In version 2, we want to have a UI for grouping notifications (but let's call this version 1 for now)

Also there was another point about notifications upon quitting fennec but I didn't understand my note on it o.o

PR #125, opened 2012-03-24
Nalexander/bug 709310 handle x weave backoff 2

ncalexan:
Reworked version of https://github.com/mozilla-services/android-sync/pull/117 with logic in GlobalSession.  Wish I had done this first, it's much simpler.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/GlobalSession.java
+++ b/src/main/java/org/mozilla/gecko/sync/GlobalSession.java
@@ -11,6 +11,7 @@ import java.net.URISyntaxException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicLong;
 
 import org.json.simple.parser.ParseException;
 import org.mozilla.gecko.sync.crypto.CryptoException;
@@ -23,6 +24,7 @@ import org.mozilla.gecko.sync.delegates.KeyUploadDelegate;
 import org.mozilla.gecko.sync.delegates.MetaGlobalDelegate;
 import org.mozilla.gecko.sync.delegates.WipeServerDelegate;
 import org.mozilla.gecko.sync.net.BaseResource;
+import org.mozilla.gecko.sync.net.HttpResponseObserver;
 import org.mozilla.gecko.sync.net.SyncResponse;
 import org.mozilla.gecko.sync.net.SyncStorageRecordRequest;
 import org.mozilla.gecko.sync.net.SyncStorageRequest;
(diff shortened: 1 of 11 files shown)


src/main/java/org/mozilla/gecko/sync/GlobalSession.java
```diff
@@ -705,4 +708,49 @@ public boolean engineIsEnabled(String engineName) throws MetaGlobalException {
...
+      return;
+    }
+
+    Logger.debug(LOG_TAG, "Observed " + responseBackoff + " millisecond backoff request.");
+    while (true) {
+      long existingBackoff = largestBackoffObserved.get();
+      if (existingBackoff >= responseBackoff) {
+        return;
+      }
+      if (largestBackoffObserved.compareAndSet(existingBackoff, responseBackoff)) {
```
rnewman, line 751: Makes you long for Clojure's concurrency primitives, doesn't it?

src/main/java/org/mozilla/gecko/sync/net/BaseResource.java
```diff
@@ -73,6 +73,8 @@
   protected HttpRequestBase request;
   public String charset = "utf-8";
 
+  protected static HttpResponseObserver httpResponseObserver = null;
```
rnewman: I think for our purposes this should be a weak reference. That is, the caller should ensure that the `HttpResponseObserver` is retained elsewhere, so that a failure to clear this due to an exception doesn't result in the object hanging around.

(Android doesn't typically clean up our process, so `BaseResource` is still loaded, and its static members will retain references. Having a `GlobalSession` dangling would be a bad thing.)

The only requirement, then, is that all of this observer's work has been done before that other reference disappears, which I think we already guarantee — `GlobalSession` does its thang before calling `handleSuccess`.

http://docs.oracle.com/javase/1.5.0/docs/api/java/lang/ref/WeakReference.html

src/main/java/org/mozilla/gecko/sync/net/HttpResponseObserver.java
```diff
@@ -0,0 +1,17 @@
...
+package org.mozilla.gecko.sync.net;
+
+import ch.boye.httpclientandroidlib.HttpResponse;
+
+public interface HttpResponseObserver {
+  /**
+   * Observe an HTTP response.
+   *
+   * @param response
+   *          The <code>HttpResponse</code> to observer.
```
rnewman: s/observer/observe

rnewman:
r+ after those.

PR #126, opened 2012-03-26
Bug 715794 - Adjust syncResult delay to implement client-count-based sync interval.

emtwo:
- Includes tests for ClientsDataDelegate functions as well.
- Should we be flagging some test in litmus to test sync intervals?

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/syncadapter/SyncAdapter.java
+++ b/src/main/java/org/mozilla/gecko/sync/syncadapter/SyncAdapter.java
@@ -51,7 +51,8 @@ public class SyncAdapter extends AbstractThreadedSyncAdapter implements GlobalSe
 
   private static final int     SHARED_PREFERENCES_MODE = 0;
   private static final int     BACKOFF_PAD_SECONDS = 5;
-  private static final int     MINIMUM_SYNC_INTERVAL_MILLISECONDS = 5 * 60 * 1000;   // 5 minutes.
+  public  static final int     MULTI_DEVICE_INTERVAL_MILLISECONDS = 5 * 60 * 1000;         // 5 minutes.
+  public  static final int     SINGLE_DEVICE_INTERVAL_MILLISECONDS = 24 * 60 * 60 * 1000;  // 24 hours.
 
   private final AccountManager mAccountManager;
   private final Context        mContext;
@@ -297,7 +298,7 @@ public class SyncAdapter extends AbstractThreadedSyncAdapter implements GlobalSe
       Log.i(LOG_TAG, "Waiting on sync monitor.");
       try {
         syncMonitor.wait();
-        long next = System.currentTimeMillis() + MINIMUM_SYNC_INTERVAL_MILLISECONDS;
+        long next = System.currentTimeMillis() + getSyncInterval();
         Log.i(LOG_TAG, "Setting minimum next sync time to " + next);
         extendEarliestNextSync(next);
       } catch (InterruptedException e) {
@@ -308,6 +309,15 @@ public class SyncAdapter extends AbstractThreadedSyncAdapter implements GlobalSe
       }
     }
  }
+
+  public int getSyncInterval() {
+    int clientsCount = this.getClientsCount();
+    if (clientsCount <= 1) {
+      return SINGLE_DEVICE_INTERVAL_MILLISECONDS;
+    }
+
+    return MULTI_DEVICE_INTERVAL_MILLISECONDS;
+  }
 
 
   /**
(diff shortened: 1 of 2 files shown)


emtwo:
Addressed Nick's comments

ncalexan:
Looks good, r=nalexander.

src/main/java/org/mozilla/gecko/sync/delegates/ClientsDataDelegate.java
```diff
@@ -4,10 +4,13 @@
...
 
+import android.accounts.Account;
+
 public interface ClientsDataDelegate {
   public String getAccountGUID();
   public String getClientName();
   public void setClientsCount(int clientsCount);
   public int getClientsCount();
   public boolean isLocalGUID(String guid);
+  public void setLocalAccount(Account localAccount);
```
rnewman: This is only ever called in test code. Don't add it to the interface; just implement it in `MockClientsDataDelegate` (or wherever) and adjust the types of call sites.

(Phrased differently: would it ever be meaningful for me to call this from real code? I think not.)

src/main/java/org/mozilla/gecko/sync/syncadapter/SyncAdapter.java
```diff
@@ -324,6 +325,15 @@ public void run() {
     }
  }
 
+  public int getSyncInterval() {
+    int clientsCount = this.getClientsCount();
+    if (clientsCount <= 1) {
+      return MINIMUM_SYNC_INTERVAL_MILLISECONDS;
+    }
+
+    return MAXIMUM_SYNC_INTERVAL_MILLISECONDS;
```
rnewman: Isn't this logic backwards? I read:

"If we have one or fewer other clients, then sync again in five minutes. If we have more clients (that care about our data), wait a whole day before syncing again".

(It would also help to document exactly what `clientsCount` is counting — _all_ clients (including us), or _other_ clients?)

src/main/java/org/mozilla/gecko/sync/syncadapter/SyncAdapter.java
```diff
@@ -56,7 +56,8 @@
 
   private static final int     SHARED_PREFERENCES_MODE = 0;
   private static final int     BACKOFF_PAD_SECONDS = 5;
-  private static final int     MINIMUM_SYNC_INTERVAL_MILLISECONDS = 5 * 60 * 1000;   // 5 minutes.
+  public  static final int     MINIMUM_SYNC_INTERVAL_MILLISECONDS = 5 * 60 * 1000;        // 5 minutes.
+  public  static final int     MAXIMUM_SYNC_INTERVAL_MILLISECONDS = 24 * 60 * 60 * 1000;  // 24 hours.
```
rnewman: The purpose of picking `MINIMUM` was to describe an absolute lower bound.

Might I suggest either `MULTI_DEVICE_INTERVAL_MILLISECONDS` and `SINGLE_DEVICE_INTERVAL_MILLISECONDS`, doing away with `MINIMUM` altogether, or adding `Math.max` to `getSyncInterval`?

test/src/org/mozilla/android/sync/test/TestClientsDataDelegate.java
```diff
@@ -0,0 +1,62 @@
+package org.mozilla.android.sync.test;
```
rnewman, line 4: License.

src/main/java/org/mozilla/gecko/sync/syncadapter/SyncAdapter.java
```diff
@@ -460,6 +470,14 @@ public synchronized int getClientsCount() {
     return Integer.parseInt(clientsCount);
   }
 
+  // Helper for testing other ClientsDataDelegate functions.
+  @Override
+  public synchronized void setLocalAccount(Account localAccount) {
+    this.localAccount = localAccount;
+    final Bundle userbundle = new Bundle();
+    mAccountManager.addAccountExplicitly(localAccount, "", userbundle);
```
rnewman: Actually, why not just directly create and delete the account before and after the test code? (`setUp` and whatever the opposite is — `tearDown`?).

emtwo:
Addressed comments.

rnewman:
It just occurred to me: Pair A Device needs to interact with this. We need to be open to syncing again really soon when we've set up a device, because the clients count will change, and we need to make sure we've uploaded everything.

Take a look at the desktop flow.

emtwo:
Filed a separate bug for adjusting sync interval after pairing devices:
https://bugzilla.mozilla.org/show_bug.cgi?id=741986

As per vidyo conversation - this bug should be fine on its own since it's unlikely for a phone to be the only device on a sync account

PR #128, opened 2012-03-27
Forms, with no AndroidBrowser* shennanigans.

ncalexan:
Much simpler to write and to unit/integration test.  Passes device functionality tests.  Cures leprosy.

### DIFF
--- a/src/main/java/org/mozilla/android/sync/test/helpers/WaitHelper.java
+++ b/src/main/java/org/mozilla/android/sync/test/helpers/WaitHelper.java
@@ -7,7 +7,7 @@ import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.TimeUnit;
 
-import android.util.Log;
+import org.mozilla.gecko.sync.Logger;
 
 /**
  * Implements waiting for asynchronous test events.
@@ -83,20 +83,12 @@ public class WaitHelper {
    */
   public static int defaultWaitTimeoutInMillis = -1;
 
-  public void info(String message) {
-    Log.i(LOG_TAG, message);
-  }
-
-  public void warn(String message, Throwable e) {
-    Log.w(LOG_TAG, message, e);
-  }
-
   public void performWait(Runnable action) throws WaitHelperError {
     this.performWait(defaultWaitTimeoutInMillis, action);
   }
 
   public void performWait(int waitTimeoutInMillis, Runnable action) throws WaitHelperError {
-    info("performWait called.");
+    Logger.debug(LOG_TAG, "performWait called.");
 
     Result result = null;
 
@@ -104,9 +96,9 @@ public class WaitHelper {
       if (action != null) {
         try {
           action.run();
-          Log.i(LOG_TAG, "Action done.");
+          Logger.debug(LOG_TAG, "Action done.");
         } catch (Exception ex) {
-          Log.i(LOG_TAG, "Performing action threw: ", ex);
+          Logger.debug(LOG_TAG, "Performing action threw: " + ex.getMessage());
           throw new InnerError(ex);
         }
       }
@@ -116,10 +108,10 @@ public class WaitHelper {
       } else {
         result = queue.poll(waitTimeoutInMillis, TimeUnit.MILLISECONDS);
       }
-      Log.i(LOG_TAG, "Got result from queue: " + result);
+      Logger.debug(LOG_TAG, "Got result from queue: " + result);
     } catch (InterruptedException e) {
       // We were interrupted.
-      info("performNotify interrupted with InterruptedException " + e);
+      Logger.debug(LOG_TAG, "performNotify interrupted with InterruptedException " + e);
       final InterruptedError interruptedError = new InterruptedError();
       interruptedError.initCause(e);
       throw interruptedError;
@@ -129,7 +121,7 @@ public class WaitHelper {
       // We timed out.
       throw new TimeoutError(waitTimeoutInMillis);
     } else if (result.error != null) {
-      warn("Notified with error.", result.error);
+      Logger.debug(LOG_TAG, "Notified with error: " + result.error.getMessage());
 
       // Rethrow any assertion with which we were notified.
       InnerError innerError = new InnerError(result.error);
@@ -140,9 +132,9 @@ public class WaitHelper {
 
   public void performNotify(final Throwable e) {
     if (e != null) {
-      warn("performNotify called with Throwable ", e);
+      Logger.debug(LOG_TAG, "performNotify called with Throwable: " + e.getMessage());
     } else {
-      info("performNotify called.");
+      Logger.debug(LOG_TAG, "performNotify called.");
     }
 
     if (!queue.offer(new Result(e))) {
--- a/src/main/java/org/mozilla/gecko/sync/repositories/android/FormHistoryRepositorySession.java
+++ b/src/main/java/org/mozilla/gecko/sync/repositories/android/FormHistoryRepositorySession.java
@@ -1,0 +1,661 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.gecko.sync.repositories.android;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.Callable;
+
+import org.mozilla.gecko.db.BrowserContract.DeletedFormHistory;
+import org.mozilla.gecko.db.BrowserContract.FormHistory;
+import org.mozilla.gecko.sync.Logger;
+import org.mozilla.gecko.sync.repositories.InactiveSessionException;
+import org.mozilla.gecko.sync.repositories.NoContentProviderException;
+import org.mozilla.gecko.sync.repositories.NoStoreDelegateException;
+import org.mozilla.gecko.sync.repositories.NullCursorException;
+import org.mozilla.gecko.sync.repositories.RecordFilter;
+import org.mozilla.gecko.sync.repositories.Repository;
+import org.mozilla.gecko.sync.repositories.StoreTrackingRepositorySession;
+import org.mozilla.gecko.sync.repositories.delegates.RepositorySessionCreationDelegate;
+import org.mozilla.gecko.sync.repositories.delegates.RepositorySessionFetchRecordsDelegate;
+import org.mozilla.gecko.sync.repositories.delegates.RepositorySessionFinishDelegate;
+import org.mozilla.gecko.sync.repositories.delegates.RepositorySessionGuidsSinceDelegate;
+import org.mozilla.gecko.sync.repositories.delegates.RepositorySessionWipeDelegate;
+import org.mozilla.gecko.sync.repositories.domain.FormHistoryRecord;
+import org.mozilla.gecko.sync.repositories.domain.Record;
+
+import android.content.ContentProviderClient;
+import android.content.ContentValues;
+import android.content.Context;
+import android.database.Cursor;
+import android.net.Uri;
+import android.os.RemoteException;
+
+public class FormHistoryRepositorySession extends
+    StoreTrackingRepositorySession {
+  public static String LOG_TAG = "FormHistoryRepoSess";
+
+  protected static Uri FORM_HISTORY_CONTENT_URI = BrowserContractHelpers.FORM_HISTORY_CONTENT_URI;
+  protected static Uri DELETED_FORM_HISTORY_CONTENT_URI = BrowserContractHelpers.DELETED_FORM_HISTORY_CONTENT_URI;
+
+  public static class FormHistoryRepository extends Repository {
+
+    @Override
+    public void createSession(RepositorySessionCreationDelegate delegate,
+                              Context context) {
+      try {
+        final FormHistoryRepositorySession session = new FormHistoryRepositorySession(this, context);
+        delegate.onSessionCreated(session);
+      } catch (Exception e) {
+        delegate.onSessionCreateFailed(e);
+      }
+    }
+  }
+
+  protected final ContentProviderClient formsProvider;
+  protected final RepoUtils.QueryHelper regularHelper;
+  protected final RepoUtils.QueryHelper deletedHelper;
+
+  /**
+   * Acquire the content provider client.
+   * <p>
+   * The caller is responsible for releasing the client.
+   *
+   * @param context The application context.
+   * @return The <code>ContentProviderClient</code>.
+   * @throws NoContentProviderException
+   */
+  public static ContentProviderClient acquireContentProvider(final Context context)
+      throws NoContentProviderException {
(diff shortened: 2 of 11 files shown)


src/main/java/org/mozilla/gecko/sync/repositories/android/FormHistoryRepositorySession.java
```diff
@@ -0,0 +1,586 @@
...
+    record.deleted = false;
+
+    record.log(LOG_TAG);
+    return record;
+  }
+
+  protected FormHistoryRecord deletedFormHistoryRecordFromCursor(Cursor cursor) {
+    String guid = RepoUtils.getStringFromCursor(cursor, BrowserContract.DeletedFormHistory.GUID);
+    String collection = "forms";
+    FormHistoryRecord record = new FormHistoryRecord(guid, collection, 0, false);
```
liuche: Look at the args for FormHistoryRecord, and set those on record creation (e.g., you're already setting the guid when you create the record, you don't need to set it again).

src/main/java/org/mozilla/android/sync/test/helpers/WaitHelper.java
```diff
@@ -104,9 +96,9 @@ public void performWait(int waitTimeoutInMillis, Runnable action) throws WaitHel
       if (action != null) {
         try {
           action.run();
-          Log.i(LOG_TAG, "Action done.");
+          Logger.debug(LOG_TAG, "Action done.");
         } catch (Exception ex) {
-          Log.i(LOG_TAG, "Performing action threw: ", ex);
+          Logger.debug(LOG_TAG, "Performing action threw: " + ex.getMessage());
```
rnewman, line 101: Could use the three-argument form here rather than concatenating.

src/main/java/org/mozilla/gecko/sync/repositories/domain/FormHistoryRecord.java
```diff
@@ -0,0 +1,137 @@
...
+      return false;
+    }
+    FormHistoryRecord other = (FormHistoryRecord) o;
+    if (!super.equalPayloads(other)) {
+      Logger.debug(LOG_TAG, "super.equalPayloads returned false.");
+      return false;
+    }
+
+    if (this.deleted) {
+      // FormHistoryRecord's are equal if they are both deleted (which
```
rnewman: No grocer's apostrophe.

src/main/java/org/mozilla/gecko/sync/repositories/domain/FormHistoryRecord.java
```diff
@@ -0,0 +1,137 @@
...
+      // they are, since super.equalPayloads is true) and have the
+      // same GUID.
+      if (other.deleted) {
+        return (RepoUtils.stringsEqual(this.guid, other.guid));
```
rnewman: Superfluous parens.

src/main/java/org/mozilla/android/sync/test/helpers/WaitHelper.java
ncalexan, line 101: Actually, I don't want the three arg form -- I'm tired of getting two stack traces on real errors.  I thought getMessage() would be short.

test/src/org/mozilla/android/sync/test/TestFormHistoryRepositorySession.java
```diff
@@ -0,0 +1,429 @@
...
+
+  public void testFetchByGuid() throws NoContentProviderException, RemoteException {
+    final FormHistoryRepositorySession session = createAndBeginSession();
+
+    insertRecords(session);
+
+    performWait(fetchRunnable(session, new String[] { record1.guid, record2.guid, record3.guid, record4.guid }, new Record[] { record1, record2, record3, record4 }));
+    performWait(fetchRunnable(session, new String[] { record1.guid, record3.guid }, new Record[] { record1, record3 }));
+    performWait(fetchRunnable(session, new String[] { record4.guid, "NON_EXISTANT_GUID?" }, new Record[] { record4 }));
+    performWait(fetchRunnable(session, new String[] { "FIRST_NON_EXISTANT_GUID", "SECOND_NON_EXISTANT_GUID?" }, new Record[] { }));
```
rnewman: Can we split lines here?

Also, "existent" not "existant".

src/main/java/org/mozilla/android/sync/test/helpers/WaitHelper.java
rnewman, line 101: wfm! :D

src/main/java/org/mozilla/gecko/sync/repositories/android/FormHistoryRepositorySession.java
```diff
@@ -0,0 +1,586 @@
...
+
+    return null;
+  }
+
+  protected Record findExistingRecordByPayload(Record rawRecord) throws RemoteException {
+    if (!rawRecord.deleted) {
+      FormHistoryRecord record = (FormHistoryRecord) rawRecord;
+      Cursor cursor = null;
+      try {
+        String where = BrowserContract.FormHistory.FIELD_NAME + " = ? AND " + BrowserContract.FormHistory.VALUE + " = ?";
```
rnewman: Line breaks, plz.

src/main/java/org/mozilla/gecko/sync/repositories/android/FormHistoryRepositorySession.java
```diff
@@ -0,0 +1,586 @@
...
+        cursor = formsProvider.query(BrowserContractHelpers.FORM_HISTORY_CONTENT_URI, null, where, new String[] { record.fieldName, record.fieldValue }, null);
```
rnewman: Line breaks or local var, plz.

src/main/java/org/mozilla/gecko/sync/repositories/android/FormHistoryRepositorySession.java
```diff
@@ -0,0 +1,586 @@
...
+        if (cursor.moveToFirst()) {
```
rnewman: Null check for cursor.

src/main/java/org/mozilla/gecko/sync/repositories/android/FormHistoryRepositorySession.java
```diff
@@ -0,0 +1,586 @@
...
+    delegate.onRecordStoreSucceeded(record);
+  }
+
+  protected Record insertNewRecord(Record rawRecord) throws RemoteException {
+    final ContentValues cv = new ContentValues();
+
+    if (rawRecord.deleted) {
+      cv.put(BrowserContract.FormHistory.GUID, rawRecord.guid);
+      // Timestamp set by content provider.
+      formsProvider.insert(BrowserContractHelpers.DELETED_FORM_HISTORY_CONTENT_URI, cv);
```
rnewman: Add a comment here that you are deliberately adding a record to the deleted items table on deletion, and why.

src/main/java/org/mozilla/gecko/sync/repositories/android/FormHistoryRepositorySession.java
```diff
@@ -0,0 +1,586 @@
...
+        }
+      }
+    }
+
+    return null;
+  }
+
+  protected void storeRecordDeletion(Record record) throws RemoteException {
+    formsProvider.delete(BrowserContractHelpers.FORM_HISTORY_CONTENT_URI, GUID_IS, new String[] { record.guid });
+    insertNewRecord(record);
```
rnewman: Note that if an error occurs half-way through here, you'll end up with neither living nor dead.

Insert the deletion first, _then_ delete from the main table.

Also, if you're going for this behavior, consider extending the Fennec `FormHistoryContentProvider` class instead. It can do all of this work (adding one, deleting the other) in a single transaction.

src/main/java/org/mozilla/gecko/sync/repositories/android/FormHistoryRepositorySession.java
rnewman: Also, be aware that you'll get an exception if you try to insert a record that's already deleted, which could happen. Either use the update-or-insert pattern here, or just catch any exception on insertion into the deleted table — it's way less important than the non-deleted table.

src/main/java/org/mozilla/gecko/sync/repositories/android/FormHistoryRepositorySession.java
```diff
@@ -0,0 +1,586 @@
...
+
+import android.content.ContentProviderClient;
+import android.content.ContentValues;
+import android.content.Context;
+import android.database.Cursor;
+import android.net.Uri;
+import android.os.RemoteException;
+
+public class FormHistoryRepositorySession extends
+    StoreTrackingRepositorySession {
```
rnewman, line 37: A linebreak here but not elsewhere?! :P

src/main/java/org/mozilla/gecko/sync/repositories/android/FormHistoryRepositorySession.java
```diff
@@ -0,0 +1,586 @@
...
+        }
+
+        ArrayList<String> guids = new ArrayList<String>();
+
+        String regularWhere = BrowserContract.FormHistory.FIRST_USED + " >= " + Long.toString(1000 * timestamp); // Microseconds.
+        String deletedWhere = BrowserContract.DeletedFormHistory.TIME_DELETED + " >= " + Long.toString(timestamp); // Milliseconds.
+
+        Cursor cur = null;
+        try {
+          cur = formsProvider.query(BrowserContractHelpers.FORM_HISTORY_CONTENT_URI, GUID_COLUMNS, regularWhere, null, null);
```
rnewman: Null check.

src/main/java/org/mozilla/gecko/sync/repositories/android/FormHistoryRepositorySession.java
```diff
@@ -0,0 +1,586 @@
...
+          delegate.onGuidsSinceFailed(e);
+          return;
+        } finally {
+          if (cur != null) {
+            cur.close();
+          }
+        }
+
+        try {
+          cur = formsProvider.query(BrowserContractHelpers.DELETED_FORM_HISTORY_CONTENT_URI, GUID_COLUMNS, deletedWhere, null, null);
```
rnewman: Null check.

src/main/java/org/mozilla/gecko/sync/repositories/android/FormHistoryRepositorySession.java
```diff
@@ -0,0 +1,586 @@
...
+        }
+
+        String guidsArray[] = guids.toArray(new String[0]);
+        delegate.onGuidsSinceSucceeded(guidsArray);
+      }
+    };
+    delegateQueue.execute(command);
+  }
+
+  protected FormHistoryRecord retrieveDuringFetch(Cursor cursor) {
```
rnewman, line 173: Null check?

src/main/java/org/mozilla/gecko/sync/repositories/android/FormHistoryRepositorySession.java
```diff
@@ -0,0 +1,586 @@
...
+    }
+  }
+
+  protected void fetchHelper(final RepositorySessionFetchRecordsDelegate delegate, final List<Callable<Cursor>> cursorCallables) {
+    if (this.storeTracker == null) {
+      throw new IllegalStateException("Store tracker not yet initialized!");
+    }
+
+    final RecordFilter filter = this.storeTracker.getFilter();
+    final long end = now();
```
rnewman: This isn't quite right.

If the first cursor callable takes a year, then the second will return records long after `end`.

I suggest two things:
1. `fetchHelper` should take a timestamp generated by the caller (i.e., wherever the cursor callables are created), and
2. those callables should use queries which have an upper bound, which is that timestamp.

That is:

```
final long sharedEnd = now();
…
// Regular callable…
BrowserContract.FormHistory.FIRST_USED + " >= " + (1000 * timestamp) + " AND " +
BrowserContract.FormHistory.LAST_USED + " <= " + (1000 * sharedEnd);

…
 
fetchHelper(delegate, sharedEnd, callableCursors);
```

src/main/java/org/mozilla/gecko/sync/repositories/android/FormHistoryRepositorySession.java
```diff
@@ -0,0 +1,586 @@
...
+  }
+
+  @Override
+  public void fetchAll(RepositorySessionFetchRecordsDelegate delegate) {
+    Logger.info(LOG_TAG, "Running fetchAll.");
+
+    Callable<Cursor> regularCallable = new Callable<Cursor>() {
+      @Override
+      public Cursor call() throws Exception {
+        return formsProvider.query(BrowserContractHelpers.FORM_HISTORY_CONTENT_URI, null, null, null, null);
```
rnewman: Even here; you want to grab all of the records for both cursors with a shared upper bound.

rnewman: Note that if the content provider exposed a URI that `UNION ALL`ed the two tables, your life would be easy here: you'd have a single query with a single cursor, and the results would be in the right format.

That's Fennec hacking, tho :)

rnewman:
Looks good; see comments :D

ncalexan:
Rebased and addressed review comments.  After second review, I will squash down.

src/main/java/org/mozilla/gecko/sync/repositories/android/FormHistoryRepositorySession.java
```diff
@@ -0,0 +1,647 @@
...
+   * @param toStore
+   *          The regular <code>Record</code> from the server.
+   * @param existingRecord
+   *          The local <code>Record</code> to replace.
+   * @throws RemoteException
+   */
+  protected void replaceExistingRecordWithRegularRecord(Record toStore, Record existingRecord)
+      throws RemoteException {
+    deleteExistingRecord(existingRecord);
+    insertNewRegularRecord(toStore);
```
rnewman: Please file a bug to be able to do this via UPDATE on the ContentProvider. No sense doing two hits on the DB in separate transactions.

test/src/org/mozilla/android/sync/test/TestFormHistoryRepositorySession.java
```diff
@@ -0,0 +1,460 @@
...
+        cur.close();
+      }
+    }
+  }
+
+  protected int numDeletedRecords(FormHistoryRepositorySession session) throws RemoteException {
+    Cursor cur = null;
+    try {
+       cur = session.getFormsProvider().query(BrowserContractHelpers.DELETED_FORM_HISTORY_CONTENT_URI, null, null, null, null);
+       return cur.getCount();
```
rnewman: Indenting wrong on these lines (and `numRegularRecords`).

ncalexan:
Final version, folded down like a lawn chair.  Used update where appropriate to save a DB operation.

ncalexan:
Batch insertions works for me (the NCE were due to database locking, which was really coming from forgetting to queue the final insert properly in storeDone).

PR #130, opened 2012-03-28
Bug 709324 - persist collection keys.

ncalexan:


rnewman:
Looking good; parts 0 and 1 are fine (but wait to land them).

The only reason I don't say "land it all, refactor for crypto version later" is because you're writing to prefs. I'd like the names to be settled and versioned before we start doing that.

ncalexan:
Closed in favour of https://github.com/mozilla-services/android-sync/pull/138

PR #132, opened 2012-03-29
Bug 738067 - Set TTLs on outbound records.

ncalexan:


src/test/java/org/mozilla/android/sync/test/TestCryptoRecord.java
```diff
@@ -50,11 +55,12 @@ public void testBaseCryptoRecordEncrypt() throws IOException, ParseException, No
...
+    String inputString = "{\"sortindex\": 131, \"ttl\": 6000, \"payload\": \"{\\\"ciphertext\\\":\\\"YJB4dr0vZEIWPirfU2FCJvfzeSLiOP5QWasol2R6ILUxdHsJWuUuvTZVhxYQfTVNou6hVV67jfAvi5Cs+bqhhQsv7icZTiZhPTiTdVGt+uuMotxauVA5OryNGVEZgCCTvT3upzhDFdDbJzVd9O3/gU/b7r/CmAHykX8bTlthlbWeZ8oz6gwHJB5tPRU15nM/m/qW1vyKIw5pw/ZwtAy630AieRehGIGDk+33PWqsfyuT4EUFY9/Ly+8JlnqzxfiBCunIfuXGdLuqTjJOxgrK8mI4wccRFEdFEnmHvh5x7fjl1ID52qumFNQl8zkB75C8XK25alXqwvRR6/AQSP+BgQ==\\\",\\\"IV\\\":\\\"v/0BFgicqYQsd70T39rraA==\\\",\\\"hmac\\\":\\\"59605ed696f6e0e6e062a03510cff742bf6b50d695c042e8372a93f4c2d37dac\\\"}\", \"id\": \"0-P9fabp9vJD\", \"modified\": 1326254123.65}";
```
rnewman: TTLs are not returned by the server, so this can go.

src/main/java/org/mozilla/gecko/sync/CryptoRecord.java
```diff
@@ -148,6 +151,9 @@ public static CryptoRecord fromJSONRecord(ExtendedJSONObject jsonRecord)
     if (jsonRecord.containsKey(KEY_SORTINDEX)) {
       record.sortIndex = jsonRecord.getLong(KEY_SORTINDEX);
     }
+    if (jsonRecord.containsKey(KEY_TTL)) {
```
rnewman, line 154: This should never be true if the record was fetched from a Sync server, so please add a comment to that effect.

PR #135, opened 2012-03-29
Handling of errors

rnewman:


liuche:
Simple enough, looks good to me. Will build with MC from this morning.

liuche:
Built with MC (minus logtag fix), verified that error is thrown in info/collections stage and the connections are closed properly. No more spamming of "Closing expired connections".

http://pastebin.mozilla.org/1544370

ncalexan:
r+ from me too.  See ticket.

PR #136, opened 2012-03-30
Bug 740247 - Follow up: HTTP 412 Error due to inconsistent server & local timestamps

emtwo:
Still needs to be tested on device + need to make sure upgrading fennec doesn't also trigger this bug, otherwise accountGUID would need to be saved in prefs rather than Account object

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/CryptoRecord.java
+++ b/src/main/java/org/mozilla/gecko/sync/CryptoRecord.java
@@ -227,8 +227,10 @@ public class CryptoRecord extends Record {
   // TODO: this only works with encrypted object, and has other limitations.
   public JSONObject toJSONObject() {
     ExtendedJSONObject o = new ExtendedJSONObject();
-    o.put(KEY_PAYLOAD, payload.toJSONString());
-    o.put(KEY_ID,      this.guid);
+    o.put(KEY_PAYLOAD,    payload.toJSONString());
+    o.put(KEY_ID,         this.guid);
+    o.put(KEY_MODIFIED,   this.lastModified);
+    o.put(KEY_COLLECTION, this.collection);
     return o.object;
   }
 
--- a/src/main/java/org/mozilla/gecko/sync/CryptoRecord.java
+++ b/src/main/java/org/mozilla/gecko/sync/CryptoRecord.java
@@ -229,8 +229,6 @@ public class CryptoRecord extends Record {
     ExtendedJSONObject o = new ExtendedJSONObject();
     o.put(KEY_PAYLOAD,    payload.toJSONString());
     o.put(KEY_ID,         this.guid);
-    o.put(KEY_MODIFIED,   this.lastModified);
-    o.put(KEY_COLLECTION, this.collection);
     return o.object;
   }
 
--- a/src/main/java/org/mozilla/gecko/sync/stage/SyncClientsEngineStage.java
+++ b/src/main/java/org/mozilla/gecko/sync/stage/SyncClientsEngineStage.java
@@ -18,6 +18,7 @@ import org.mozilla.gecko.sync.GlobalSession;
 import org.mozilla.gecko.sync.HTTPFailureException;
 import org.mozilla.gecko.sync.Logger;
 import org.mozilla.gecko.sync.NoCollectionKeysSetException;
+import org.mozilla.gecko.sync.Utils;
 import org.mozilla.gecko.sync.crypto.CryptoException;
 import org.mozilla.gecko.sync.crypto.KeyBundle;
 import org.mozilla.gecko.sync.delegates.ClientsDataDelegate;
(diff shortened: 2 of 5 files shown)


emtwo:
Tested on device with the logging and functionality is as expected.

src/main/java/org/mozilla/gecko/sync/CryptoRecord.java
```diff
@@ -227,8 +227,10 @@ protected void initFromPayload(ExtendedJSONObject payload) {
   // TODO: this only works with encrypted object, and has other limitations.
   public JSONObject toJSONObject() {
     ExtendedJSONObject o = new ExtendedJSONObject();
-    o.put(KEY_PAYLOAD, payload.toJSONString());
-    o.put(KEY_ID,      this.guid);
+    o.put(KEY_PAYLOAD,    payload.toJSONString());
+    o.put(KEY_ID,         this.guid);
+    o.put(KEY_MODIFIED,   this.lastModified);
+    o.put(KEY_COLLECTION, this.collection);
```
rnewman: Can you explain why you added these two lines?

src/main/java/org/mozilla/gecko/sync/stage/SyncClientsEngineStage.java
```diff
@@ -164,18 +180,34 @@ public String credentials() {
...
     public void handleRequestSuccess(SyncStorageResponse response) {
       Logger.debug(LOG_TAG, "Upload succeeded.");
-      commandsProcessedShouldUpload = false;
-      uploadAttemptsCount.set(0);
-      session.config.persistServerClientRecordTimestamp(response.normalizedWeaveTimestamp());
+      try {
+        commandsProcessedShouldUpload = false;
+        uploadAttemptsCount.set(0);
+
+        long timestamp = Utils.decimalSecondsToMilliseconds(response.body().toString());
```
rnewman: `response.body()` returns a `String`, so you can lose the `.toString()`.

src/main/java/org/mozilla/gecko/sync/stage/SyncClientsEngineStage.java
```diff
@@ -164,18 +180,34 @@ public String credentials() {
...
+        session.config.persistServerClientRecordTimestamp(timestamp);
+        BaseResource.consumeEntity(response);
 
-      BaseResource.consumeEntity(response);
+        Logger.info(LOG_TAG, "Timestamp from body is: " + timestamp);
+        Logger.info(LOG_TAG, "Timestamp from header is: " + response.normalizedWeaveTimestamp());
```
rnewman: s/info/debug, I think.

src/main/java/org/mozilla/gecko/sync/stage/SyncClientsEngineStage.java
```diff
@@ -164,18 +180,34 @@ public String credentials() {
...
+      } catch (Exception e) {
+        e.printStackTrace();
```
rnewman: This can go.

src/test/java/org/mozilla/android/sync/net/test/TestClientsEngineStage.java
```diff
@@ -89,6 +96,12 @@ public void setup() {
     }
   }
 
+  @After
+  public void teardown() {
+    stubUpload = false;
+    ((MockClientsDatabaseAccessor)db).resetVars();
```
rnewman: Space after `(Cast)`.

src/test/java/org/mozilla/android/sync/net/test/TestClientsEngineStage.java
```diff
@@ -168,22 +185,50 @@ public void handleWBO(CryptoRecord record) {
...
-    public MockSuccessClientUploadDelegate(HTTPServerTestHelper data) {
+  public class TestHandleWBODownloadDelegate extends TestClientDownloadDelegate {
+    public TestHandleWBODownloadDelegate(HTTPServerTestHelper data) {
       super(data);
     }
 
     @Override
     public void handleRequestFailure(SyncStorageResponse response) {
       super.handleRequestFailure(response);
+      assertTrue(((MockClientsDatabaseAccessor)db).closed);
```
rnewman: Space. (Same throughout.)

src/main/java/org/mozilla/gecko/sync/CryptoRecord.java
emtwo: The collection line I added in exchange for the following lines I removed from UploadMockServer in TestClientsEngineStage:

```
// Note: collection is not saved in CryptoRecord.toJSONObject() upon upload.
// So its value is null and is set here so ClientRecord.equals() may be used.
uploadedRecord.collection = lastComputedLocalClientRecord.collection;
```

Simply because it seemed neater to me. I assume the reason it was not there before was because we are uploading to a clients collection and don't need to write it in every record on the server? If so, I'm happy to remove it.

The lastModified line I added to help test `handleWBO(CryptoRecord record)` in the download delegate. The delegate accesses a downloaded record's `lastModified` field so I use `toJSONObject()` in my `DownloadLocalRecordMockServer` to pass down a record with a lastModified field.

Perhaps this was not the right place to make this addition? Should I instead be creating a "modified" key/value pair from within the MockServer?

emtwo: tl;dr - I can remove these lines if they aren't appropriate here & make the appropriate changes in the tests

emtwo:
Addressed comments.

PR #137, opened 2012-03-30
Bug 709385 - Passwords repository layer

liuche:
Passwords and deletions are syncing, as in, they appear in the database, but they do not actually autocomplete - likely needs form history to autocomplete.

Passes tests, works as expected in smoketesting.

// FIXED: Find record query doesn't seem to work correctly, would like feedback on findExistingRecord "where" building.


src/main/java/org/mozilla/gecko/sync/repositories/android/PasswordsRepositorySession.java
```diff
@@ -0,0 +1,685 @@
...
+          record.hostname,
+          record.formSubmitURL,
+          // TODO: Bug 738347 - SQLiteBridge does not check for nulls in ContentValues.
+          // record.httpRealm,
+          record.usernameField,
+          record.passwordField
+      };
+    try {
+      cursor = safeQuery(passwordsProvider, getUriDeleted(false), ".findRecord", null, dataWhere, whereArgs, null);
+      if (cursor != null && cursor.moveToFirst()) {
```
ncalexan, line 564: Cursor won't be null here.

src/main/java/org/mozilla/gecko/sync/repositories/android/PasswordsRepositorySession.java
```diff
@@ -0,0 +1,685 @@
...
+        while (!cursor.isAfterLast()) {
+          foundRecord = passwordRecordFromCursorDeleted(cursor, false);
+          if (record.encryptedUsername.equals(foundRecord.encryptedUsername)) {
```
ncalexan, line 567: This seems strange.  I know this was causing trouble, but this should be done by the db.  Unless you don't want to send the encryptedUsername in as an arg?

src/main/java/org/mozilla/gecko/sync/repositories/android/PasswordsRepositorySession.java
```diff
@@ -0,0 +1,685 @@
...
+  private PasswordRecord passwordRecordFromCursorDeleted(Cursor cur, boolean deleted) {
+    if (cur == null || !cur.moveToFirst()) {
+      return null;
+    }
+    PasswordRecord rec;
+    if (deleted) {
+      String guid = RepoUtils.getStringFromCursor(cur, DeletedColumns.GUID);
+      long lastModified = RepoUtils.getLongFromCursor(cur, DeletedColumns.TIME_DELETED);
+      rec = new PasswordRecord(guid, COLLECTION, lastModified, true);
+      rec.androidID = RepoUtils.getLongFromCursor(cur, DeletedColumns.ID);
```
ncalexan, line 612: nit: early return here, de-indent later.

src/main/java/org/mozilla/gecko/sync/repositories/android/PasswordsRepositorySession.java
```diff
@@ -0,0 +1,685 @@
...
+    cv.put(BrowserContract.Passwords.ENCRYPTED_PASSWORD, rec.encryptedPassword);
+
+    cv.put(BrowserContract.Passwords.TIME_CREATED,          rec.timeCreated);
+    cv.put(BrowserContract.Passwords.TIME_LAST_USED,        rec.timeLastUsed);
+    cv.put(BrowserContract.Passwords.TIME_PASSWORD_CHANGED, rec.timePasswordChanged);
+    cv.put(BrowserContract.Passwords.TIMES_USED,            rec.timesUsed);
+    return cv;
+  }
+
+  // shameless copying + tweaking.
```
ncalexan, line 664: Now that forms has landed, this is all in RepoUtils.

src/main/java/org/mozilla/gecko/sync/repositories/domain/PasswordRecord.java
```diff
@@ -100,20 +133,48 @@ public boolean equalPayloads(Object o) {
...
         && RepoUtils.stringsEqual(this.usernameField, other.usernameField)
         && RepoUtils.stringsEqual(this.passwordField, other.passwordField)
-        && RepoUtils.stringsEqual(this.encType, other.encType)
-        && (this.timeLastUsed == other.timeLastUsed)
-        && (this.timesUsed == other.timesUsed);
+        && RepoUtils.stringsEqual(this.encryptedUsername, other.encryptedUsername)
+        && RepoUtils.stringsEqual(this.encryptedPassword, other.encryptedPassword);
+        // Desktop sync never sets timeCreated so this isn't relevant for sync records.
+        // && (this.timeCreated == other.timeCreated)
+//        && (this.timeLastUsed == other.timeLastUsed)
```
ncalexan, line 157: Nit: cleaning.

src/main/java/org/mozilla/gecko/sync/repositories/domain/PasswordRecordFactory.java
```diff
@@ -0,0 +1,25 @@
...
+import org.mozilla.gecko.sync.repositories.RecordFactory;
+
+/**
+ * Turns CryptoRecords into PasswordRecords.
+ *
+ * @author liuche
+ *
+ */
+
+public class PasswordRecordFactory extends RecordFactory {
```
ncalexan, line 17: Nit: For forms and tabs, we moved this inline.  Saves a file.

src/main/java/org/mozilla/gecko/sync/repositories/InactiveSessionException.java
```diff
@@ -43,6 +43,9 @@
 
   private static final long serialVersionUID = 537241160815940991L;
 
+  public InactiveSessionException() {
```
ncalexan, line 46: nit: just send in a null exception. (same below)

src/main/java/org/mozilla/gecko/sync/repositories/PasswordsRepository.java
```diff
@@ -0,0 +1,29 @@
...
+
+package org.mozilla.gecko.sync.repositories;
+
+import org.mozilla.gecko.sync.repositories.android.PasswordsRepositorySession;
+import org.mozilla.gecko.sync.repositories.delegates.RepositorySessionCreationDelegate;
+
+import android.content.Context;
+
+/**
+ * Shared interface for repositories that consume and produce
```
ncalexan, line 13: nit: it's not an interface.  For forms and tabs, this is an inner class -- I would prefer it be inner to your PasswordsRepositorySession.

src/main/java/org/mozilla/gecko/sync/repositories/android/PasswordsRepositorySession.java
liuche, line 567: The username/password is encrypted in the DB, and in the repoSession, we don't have access to the encryption keys. So the only way is to filter through the records returned by a more general query.

I'll add a comment.

src/main/java/org/mozilla/gecko/sync/repositories/RepositorySession.java
```diff
@@ -84,6 +88,15 @@ public RepositorySession(Repository repository) {
   public abstract void fetch(String[] guids, RepositorySessionFetchRecordsDelegate delegate) throws InactiveSessionException;
   public abstract void fetchAll(RepositorySessionFetchRecordsDelegate delegate);
 
+  protected ContentProviderClient getContentProvider(final Context context, final Uri uri) throws NoContentProviderException {
```
ncalexan, line 91: This seems in the wrong place.  There are repository sessions with no CP anywhere behind them.

src/main/java/org/mozilla/gecko/sync/repositories/android/PasswordsRepositorySession.java
```diff
@@ -0,0 +1,685 @@
...
+  private final Context context;
+
+  public PasswordsRepositorySession(Repository repository, Context context) {
+    super(repository);
+    this.context = context;
+    passwordsProvider = context.getContentResolver().acquireContentProviderClient(BrowserContract.PASSWORDS_AUTHORITY_URI);
+  }
+
+  @Override
+  public void guidsSince(final long timestamp,
```
ncalexan, line 52: If you wanted, you could lift what's in forms for doing this multiple cursor fetch with Callables, but your sequencing looks good.

src/main/java/org/mozilla/gecko/sync/repositories/android/PasswordsRepositorySession.java
```diff
@@ -0,0 +1,685 @@
...
+
+        } catch (NullCursorException e) {
+          Logger.error(LOG_TAG, "Null cursor in fetch.");
+          delegate.onFetchFailed(e, null);
+        } catch (Exception e) {
+          Logger.error(LOG_TAG, "Exception in fetch.");
+          delegate.onFetchFailed(e, null);
+        } finally {
+          Logger.info(LOG_TAG, "Closing cursor after fetch.");
+          cursor.close();
```
ncalexan, line 146: Check for null cursor.

src/main/java/org/mozilla/gecko/sync/repositories/domain/PasswordRecord.java
```diff
@@ -100,20 +133,48 @@ public boolean equalPayloads(Object o) {
     if (o == null || !(o instanceof PasswordRecord)) {
       return false;
     }
+
     PasswordRecord other = (PasswordRecord) o;
+    Logger.debug("PasswordRecord", "thisRecord:" + this.toString());
+    Logger.debug("PasswordRecord", "otherRecord:" + o.toString());
+
     if (!super.equalPayloads(other)) {
```
ncalexan, line 141: Have you tested that your `PasswordRecord.equals` handles all permuations of deleted?  I know I needed to handle this case in forms -- if both were deleted, too much was tested.

src/main/java/org/mozilla/gecko/sync/repositories/android/PasswordsRepositorySession.java
```diff
@@ -0,0 +1,685 @@
...
+    return newRecord;
+  }
+
+  private void updateTimes(Record record) {
+    PasswordRecord passwordRecord = (PasswordRecord) record;
+    passwordRecord.timePasswordChanged = now();
+  }
+
+  // Helper Functions.
+  private Uri getUriDeleted(boolean deleted) {
```
ncalexan, line 430: This is purely stylistic, but I am not a fan of this `deleted` flag everywhere.  AFAICT, the flag is always static.  I would really rather have two functions with descriptive names and assumptions in the docstrings than have an if/then block, but that is my personal preference.  rnewman?  marina?

test/src/org/mozilla/android/sync/test/helpers/PasswordHelpers.java
```diff
@@ -73,11 +81,13 @@ public static PasswordRecord createPassword5() {
...
     rec.httpRealm = "httpRealm5";
-    rec.password ="54325";
+    rec.encryptedPassword ="54325";
     rec.passwordField = "box.pass.field5";
+    rec.timeCreated = 101000000000L;
     rec.timeLastUsed = 123412352555L;
+    rec.timePasswordChanged = 111111111111L;
     rec.timesUsed = 5L;
-    rec.username = "jvoll5";
+    rec.encryptedUsername = "jvoll5";
```
ncalexan, line 90: Looking really good!  I just did a little fly-by review.

src/main/java/org/mozilla/gecko/sync/repositories/android/PasswordsRepositorySession.java
ncalexan, line 567: I don't think you need to decrypt, just ask the db to check for `Passwords.ENCRYPTED_USERNAME + " = ? AND "` + with record.encryptedUsername as an argument.  But you've been working on this, and that's the obvious solution, so please document why it doesn't work.

liuche:
Fixes and rebase in pull request #150:

https://github.com/mozilla-services/android-sync/pull/150

src/main/java/org/mozilla/gecko/sync/repositories/android/PasswordsRepositorySession.java
liuche, line 567: I think the confusion in these comments might be due to naming. The encrypted username is not encrypted with the sync key. (I believe the encryption refers to encryption by the Master Password, which does not exist ATM.)

Since the username/password is not encrypted, and we have no way of encrypting it, the only way to find a matching record is to filter through the unencrypted records returned by the query.

PR #139, opened 2012-03-30
Bug 725912: Sync key validation + tests.

liuche:
Note: there is a string in this commit, informing the user that the entered Sync Key format is incorrect.


rnewman:
Let's also add strings for all of the password error/change stuff. One and done.

liuche:
Rebased to:

https://github.com/mozilla-services/android-sync/pull/141

PR #142, opened 2012-04-02
Bug 738515: Null credentials in setup.

liuche:
Added checks for null credentials at the end of setup. Also added check for 0-length credentials, because that's what's sent if you complete setup and don't have credentials.


liuche:
Also, nothing we can do about desktop thinking we have successfully set up sync. This seems to also be the same in desktop-to-desktop sync setup. (If you delete the credentials partway through setup, both desktop Firefoxes think setup has completed.)

PR #144, opened 2012-04-03
Bug 730142 - download batching

ncalexan:
Implemented pulling all guids (no limit at the moment) and filling over time.  Final patch includes humorously low limits for testing.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/middleware/Crypto5MiddlewareRepositorySession.java
+++ b/src/main/java/org/mozilla/gecko/sync/middleware/Crypto5MiddlewareRepositorySession.java
@@ -110,18 +110,6 @@ public class Crypto5MiddlewareRepositorySession extends MiddlewareRepositorySess
     }
 
     @Override
-    public void onFetchSucceeded(Record[] records, long fetchEnd) {
-      for (Record record : records) {
-        try {
-          this.onFetchedRecord(record);
-        } catch (Exception e) {
-          this.onFetchFailed(e, record);
-        }
-      }
-      this.onFetchCompleted(fetchEnd);
-    }
-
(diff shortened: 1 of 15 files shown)


ncalexan:
Oh yes, there are live tests here too, which maybe should not be merged.  Some pieces are not tested live, but it's hard to really get the crucial pieces without... you know... a server on the other end.

rnewman:
I would like to see as much of the filler code as possible factored out to use composition rather than inheritance.

That is, there's an object that knows how to store GUIDs for later use, and give you more when you ask for them.

A `FillingServer11Repository`, then, is simply one that does a first sync via a GUID grab, augments fetches from the GUID store, and hook in resets etc. appropriately.

But more broadly I'd like you to consider whether this is part of the `Repository` at all. I'd contend that it's part of the `Synchronizer`: that is, this is a kind of `Synchronizer` that, on first sync, doesn't do fetchSince-store, fetchSince-store, but instead does guidsSince-fetch-store, fetchSince-store.

That way I can apply this to any `Repository` I like, just by pairing it with a different `Synchronizer`. It'll set up an appropriate container to stick in the way, and use that as the source for the `RecordsChannel`.

This also avoids the uncomfortable alteration of `fetchSince`'s semantics — that is, using _this_ approach, you'll be returning records in `fetchSince` that were modified before the requested timestamp! That ain't right.

rnewman:
Not a blocker, so this pull request languishes for a couple weeks.

rnewman:
Nick: please rebase.

ncalexan:
There are salvageable parts of this (mostly things from `Server11Repository`) but the main download batching mechanism needs to be rewritten, so I'm closing this pull req.

PR #148, opened 2012-04-05
Bug 742700 - Null URI crash in canceling J-PAKE.

liuche:
Do not try to delete channel if user quits in GetChannelStage. Also added a safeguard check in DeleteChannel for null Uri, even though we shouldn't reach this part of the code anymore, if we have a null Uri.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/jpake/JPakeClient.java
+++ b/src/main/java/org/mozilla/gecko/sync/jpake/JPakeClient.java
@@ -72,7 +72,7 @@ public class JPakeClient {
   public int                 jpakePollInterval;
   public int                 jpakeMaxTries;
   public String              channel;
-  public String              channelUrl;
+  public volatile String     channelUrl;
 
   // J-PAKE session data.
   public KeyBundle           myKeyBundle;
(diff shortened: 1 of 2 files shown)


src/main/java/org/mozilla/gecko/sync/jpake/JPakeClient.java
```diff
@@ -240,7 +240,8 @@ public void abort(String reason) {
     finished = true;
     if (Constants.JPAKE_ERROR_CHANNEL.equals(reason) ||
         Constants.JPAKE_ERROR_NETWORK.equals(reason) ||
-        Constants.JPAKE_ERROR_NODATA.equals(reason)) {
+        Constants.JPAKE_ERROR_NODATA.equals(reason) ||
+        currentStage instanceof GetChannelStage) {
```
ncalexan: Why not test if `channelUrl == null` here, rather than below, and don't add the `currentStage` instance variable at all?

liuche: That actually makes a lot of sense. Incorporated in the newest commit.

rnewman:
If I understand the situation correctly, there are two threads at play here: the UI thread (which is the one calling `abort()` from `onPause()`) and the J-PAKE runnable (e.g., calling into `finishResume` from `onResume` on the `ThreadPool`, which kicks off the sequence of J-PAKE stages.

That means you can get concurrent access to `channelUrl`, and also that this situation can arise because the channel URL is being fetched at the same time as the session is being aborted.

This is a long-winded way of saying that this is a partial and not-thread-safe solution to the problem.

liuche:
Made channelUrl a volatile variable. This may still cause channel leak if an abort from the UI stops sync setup, but comments mention this case, and the server will delete the channel after a period of time anyways.

PR #149, opened 2012-04-05
Bug 741960 - fuzz backoff

ncalexan:
This was a weird ticket.  The actual fuzzing is trivial, but I don't think I hit the separation you (rnewman) intended.  I committed in tiny pieces so you could mix and match as you see fit.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/syncadapter/SyncAdapter.java
+++ b/src/main/java/org/mozilla/gecko/sync/syncadapter/SyncAdapter.java
@@ -8,6 +8,7 @@ import java.io.IOException;
 import java.net.URI;
 import java.security.NoSuchAlgorithmException;
 import java.util.concurrent.TimeUnit;
+
 import org.json.simple.parser.ParseException;
 import org.mozilla.gecko.sync.AlreadySyncingException;
 import org.mozilla.gecko.sync.GlobalConstants;
@@ -41,7 +42,6 @@ import android.database.sqlite.SQLiteConstraintException;
 import android.database.sqlite.SQLiteException;
 import android.os.Bundle;
 import android.os.Handler;
-import android.util.Log;
 
 public class SyncAdapter extends AbstractThreadedSyncAdapter implements GlobalSessionCallback, ClientsDataDelegate {
   private static final String  LOG_TAG = "SyncAdapter";
@@ -61,7 +61,7 @@ public class SyncAdapter extends AbstractThreadedSyncAdapter implements GlobalSe
   public SyncAdapter(Context context, boolean autoInitialize) {
     super(context, autoInitialize);
     mContext = context;
-    Log.d(LOG_TAG, "AccountManager.get(" + mContext + ")");
+    Logger.debug(LOG_TAG, "AccountManager.get(" + mContext + ")");
     mAccountManager = AccountManager.get(context);
   }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/mozilla/gecko/sync/GlobalSession.java
```diff
@@ -302,7 +302,7 @@ public void start() throws AlreadySyncingException {
    */
   protected void restart() throws AlreadySyncingException {
     this.currentState = GlobalSyncStage.Stage.idle;
-    if (callback.shouldBackOff()) {
+    if (callback.shouldBackoff()) {
```
rnewman: Actually two words — "should I back off? My backoff is set to 10", just like "can I log in? My login credentials are".

src/main/java/org/mozilla/gecko/sync/syncadapter/SyncAdapter.java
```diff
@@ -72,24 +72,20 @@ private SharedPreferences getGlobalPrefs() {
   /**
    * Backoff.
    */
-  public synchronized long getEarliestNextSync() {
-    SharedPreferences sharedPreferences = getGlobalPrefs();
-    return sharedPreferences.getLong(PREFS_EARLIEST_NEXT_SYNC, 0);
+  protected synchronized long getEarliestNextSync() {
+    return getGlobalPrefs().getLong(PREFS_EARLIEST_NEXT_SYNC, 0);
```
rnewman: This is really a per-server thing. You could store that per server, or you can store it per account and make sure you delete it when you get a new cluster assignment.

src/main/java/org/mozilla/gecko/sync/syncadapter/SyncAdapter.java
```diff
@@ -201,7 +184,26 @@ public boolean shouldBackOff() {
...
+      this.extendEarliestNextSync(System.currentTimeMillis() + backoff);
+    }
+  }
+
+  protected void setAndroidBackoff(SyncResult syncResult) {
+    long now  = System.currentTimeMillis();
+    long next = getEarliestNextSync();
+    long seconds = Math.max((next - now) / 1000, 0) + BACKOFF_PAD_SECONDS;
+    Logger.info(LOG_TAG, "Asking Android to wait another " + seconds + " seconds.");
+    syncResult.delayUntil = seconds;
```
rnewman: This is where that separation comes in.

If we ask Android to not call us for 24 hours, and then we pair a new device, we won't have the opportunity to sync again when we should.

So there needs to be a `Math.min` in here with some "try again" interval (still padded, of course).

5 minutes seems about right for now — that is, the smallest interval at which we do want to sync.

The exception to this is a backoff/retry-after request from the server. We absolutely do want to honor that. This might require us to split `getEarliestNextSync` and `getEarliestNextPerformSyncCall` — the former is `seconds` here, and the latter is the smallest of 5 minutes and the requested backoff, unless the backoff is server-requested.

I'm open to how you want to actually effect this, but this `delayUntil` phrase is the tip of the spear.

ncalexan:
Let's see if this is closer to what we want.

ncalexan:
Rebased, refactored to reduce name churn.

rnewman:
Not a blocker, so this pull request languishes for a couple weeks.

rnewman:
Nick: please rebase.

PR #151, opened 2012-04-06
Bug 740193 - Send tab to device: push tabs to other devices

emtwo:
Still rough, but you can see the code for the UI so far if anyone is interested in reviewing.

The UI is missing a couple of small things still:
- I want to redirect users to the sync setup page if they try to share via firefox but aren't setup (right now this will just show a blank Send Tab to Devices page)
- Right now when the local record is on the server, it's downloaded & saved to the db then displayed in the send-tab list. Probably shouldn't be saving our own record in the db in the first place.
- Code should probably be moved to a separate package for send tab to device.

Other than that, sendClickHandler() has access to the uri, title, and guids to send to and the next step is to validate and save this information in a table in the clients db


ncalexan:
Have some review comments in patch form.  Could we:
- have checkboxes on the left?  On my tablet, the text and the checkbox are miles apart, which is odd.
- make it so the checkbox state doesn't get wiped on screen orientation change?

Otherwise, this is looking good so far.

ncalexan:
Can the state of the checkboxes be stored in the UI widgets rather than in our own `boolean[]` state?

emtwo:
Potentially... I'll take a look at it once I've got the backend working since I'll be making some other UI changes as well.

emtwo:
Closing this pull request and opening another more fresh one.

PR #152, opened 2012-04-10
Bug 714657 - purge stored prefs

ncalexan:
I got rid of "sync.prefs.global" -- don't understand why it was there in the first place, and some decidely not-global state (stale clusterURL flag) was being written there.  Now we just take care to get our prefs in place for the correct account before testing for earliest next sync.

This will probably conflict with https://github.com/mozilla-services/android-sync/pull/138, which is languishing...

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/Utils.java
+++ b/src/main/java/org/mozilla/gecko/sync/Utils.java
@@ -54,6 +54,7 @@ import java.util.TreeMap;
 import org.json.simple.JSONArray;
 import org.mozilla.apache.commons.codec.binary.Base32;
 import org.mozilla.apache.commons.codec.binary.Base64;
+import org.mozilla.gecko.sync.crypto.KeyBundle;
 
 import android.content.Context;
 import android.content.SharedPreferences;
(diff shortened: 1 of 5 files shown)


src/main/java/org/mozilla/gecko/sync/Utils.java
```diff
@@ -215,17 +216,53 @@ public static String sha1Base32(String utf8)
...
 
+  /**
+   * Return the persisted preferences path for a specific account.
+   *
+   * @param username Plain username, e.g., "user@mozilla.com".
+   */
   public static String getPrefsPath(String username, String serverURL)
     throws NoSuchAlgorithmException, UnsupportedEncodingException {
-    return "sync.prefs." + sha1Base32(serverURL + ":" + username);
+    return "sync.prefs." + sha1Base32(serverURL + ":" + KeyBundle.usernameFromAccount(username));
```
rnewman, line 226: No need for this change; taking the plain username ensures that you only get the right branch if you know the real username, and after this we're just going to hash it twice.

src/main/java/org/mozilla/gecko/sync/syncadapter/SyncAdapter.java
```diff
@@ -65,25 +64,30 @@ public SyncAdapter(Context context, boolean autoInitialize) {
...
+  private SharedPreferences getPrefs() {
+    return prefs;
   }
 
   /**
    * Backoff.
    */
   public synchronized long getEarliestNextSync() {
-    SharedPreferences sharedPreferences = getGlobalPrefs();
+    SharedPreferences sharedPreferences = getPrefs();
```
rnewman, line 80: Make sure this isn't the path for backoff. Backoff _is_ shared between accounts on the same server.

src/main/java/org/mozilla/gecko/sync/syncadapter/SyncAdapter.java
```diff
@@ -65,25 +64,30 @@ public SyncAdapter(Context context, boolean autoInitialize) {
     mAccountManager = AccountManager.get(context);
   }
 
-  private SharedPreferences getGlobalPrefs() {
-    return mContext.getSharedPreferences("sync.prefs.global", SHARED_PREFERENCES_MODE);
```
rnewman, line 69: You can't make this change without first implementing a prefs versioning and migration scheme, or verifying 100% that you're not going to break existing accounts.

And are you totally confident that this branch doesn't add value — such as the ability to wipe all global prefs, or all prefs for all accounts, or whatnot, independent of others?

We're eventually (and I mean "when I write the resetting code") going to move to a versioned `account.product.profile` structure, so having a known place to put stuff that doesn't get blown away seems like a good idea to me.

rnewman:
Err...

Bug https://bugzilla.mozilla.org/show_bug.cgi?id=714647 nor, --, ---, general, NEW, Add rooters to get rooting analysis to pass sunspider/v8bench

ncalexan:
Commit message was wrong in some places, sorry: https://bugzilla.mozilla.org/show_bug.cgi?id=714657.

I was hoping to do some prep work so resetting would be easier, but I think I'll find other things to work on.  Feel free to close this.

rnewman:
Closing for now.

PR #153, opened 2012-04-10
Bug 724739, Bug 724740

rnewman:
Test to follow, but it passes existing tests!

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksDataAccessor.java
+++ b/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksDataAccessor.java
@@ -5,7 +5,9 @@
 package org.mozilla.gecko.sync.repositories.android;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
+import java.util.Map;
 
 import org.json.simple.JSONArray;
 import org.mozilla.gecko.db.BrowserContract;
@@ -105,6 +107,21 @@ public class AndroidBrowserBookmarksDataAccessor extends AndroidBrowserRepositor
     Logger.debug(LOG_TAG, "Updating positions for " + size + " items.");
     String[] args = childArray.toArray(new String[size]);
     return context.getContentResolver().update(getPositionsUri(), new ContentValues(), null, args);
+  }
+
+  public int bumpModifiedByGUID(Collection<String> ids, long modified) {
+    final int size = ids.size();
+    if (size == 0) {
+      return 0;
+    }
+
+    Logger.debug(LOG_TAG, "Bumping modified for " + size + " items to " + modified);
+    String where = RepoUtils.computeSQLInClause(size, BrowserContract.Bookmarks.GUID);
+    String[] selectionArgs = ids.toArray(new String[size]);
+    ContentValues values = new ContentValues();
+    values.put(BrowserContract.Bookmarks.DATE_MODIFIED, modified);
+
+    return context.getContentResolver().update(getUri(), values, where, selectionArgs);
   }
 
   /**
--- a/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksDataAccessor.java
+++ b/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksDataAccessor.java
@@ -79,10 +79,13 @@ public class AndroidBrowserBookmarksDataAccessor extends AndroidBrowserRepositor
     context.getContentResolver().delete(uri, EXCLUDE_SPECIAL_GUIDS_WHERE_CLAUSE, null);
   }
 
+  private String[] GUID_AND_ID = new String[] { BrowserContract.Bookmarks.GUID,
+                                                BrowserContract.Bookmarks._ID };
+
   protected Cursor getGuidsIDsForFolders() throws NullCursorException {
     // Exclude "places" and "tags", in case they've ended up in the DB.
     String where = BOOKMARK_IS_FOLDER + " AND " + GUID_NOT_TAGS_OR_PLACES;
-    return queryHelper.safeQuery(".getGuidsIDsForFolders", null, where, null, null);
+    return queryHelper.safeQuery(".getGuidsIDsForFolders", GUID_AND_ID, where, null, null);
   }
 
   /**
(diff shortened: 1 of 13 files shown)


src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
```diff
@@ -689,35 +714,32 @@ protected void storeRecordDeletion(final Record record) {
...
     }
     final BookmarkRecord bookmarkRecord = (BookmarkRecord) record;
-    if (bookmarkRecord.isFolder()) {
-      Logger.debug(LOG_TAG, "Deleting folder. Ensuring consistency of children. TODO: Bug 724470.");
-      handleFolderDeletion(bookmarkRecord);
-      return;
+    deletionManager.deleteRecord(bookmarkRecord);
+  }
+
+  protected void untrackGUIDs(Collection<String> guids, long modified) {
```
ncalexan: This is gnarly.  Replace `untrackRecord` up the hiearchy with `untrackRecords(String[] guids)`.

src/main/java/org/mozilla/gecko/sync/repositories/android/BookmarksDeletionManager.java
```diff
@@ -0,0 +1,234 @@
...
+ * from within a store runnable.
+ *
+ */
+public class BookmarksDeletionManager {
+  private static final String LOG_TAG = "BookmarkDelete";
+
+  private final AndroidBrowserBookmarksDataAccessor dataAccessor;
+  private RepositorySessionStoreDelegate delegate;
+
+  private final int flushThreshold;
```
ncalexan, line 55: Somewhere, this needs a comment.  Here, constructor, or in the session class is fine.

src/main/java/org/mozilla/gecko/sync/repositories/android/BookmarksDeletionManager.java
```diff
@@ -0,0 +1,234 @@
...
+
+  private final HashSet<String> folders    = new HashSet<String>();
+  private final HashSet<String> nonFolders = new HashSet<String>();
+  private int folderCount = 0;
```
ncalexan: When is `folderCount != folders.size()`?  I'd much rather this be computed by the data structure rather than kept as independent state.

src/main/java/org/mozilla/gecko/sync/repositories/android/BookmarksDeletionManager.java
```diff
@@ -0,0 +1,234 @@
...
+      if (folders.add(r.guid)) {
+        ++folderCount;
+      }
+      if (!folders.contains(r.parentID)) {
+        // We're not going to delete its parent; will need to bump it.
+        folderParents.add(r.parentID);
+      }
+
+      nonFolderParents.remove(r.guid);
+      folderParents.remove(r.guid);
```
ncalexan, line 106: nit: prefer return over else.

src/main/java/org/mozilla/gecko/sync/repositories/android/BookmarksDeletionManager.java
```diff
@@ -0,0 +1,234 @@
...
+          deleteNonFolders();
+        }
+      }
+    }
+  }
+
+  /**
+   * Flush deletions that can be easily taken care of right now.
+   */
+  public void incrementalFlush() {
```
ncalexan, line 125: Never called, get rid of it.

src/main/java/org/mozilla/gecko/sync/repositories/android/BookmarksDeletionManager.java
```diff
@@ -0,0 +1,234 @@
...
+  /**
+   * Apply all pending deletions and reset state for the next batch of stores.
+   *
+   * @param orphanDestination the ID of the folder to which orphaned children
+   *                          should be moved.
+   *
+   * @throws NullCursorException
+   * @return a set of IDs to untrack. Will not be null.
+   */
+  public Set<String> completeFlush(long orphanDestination, long now) throws NullCursorException {
```
ncalexan: nit: `completeFlush` is ambiguous.  Does that mean "complete a flush that is progress" or "do a complete flush"?

src/main/java/org/mozilla/gecko/sync/repositories/android/BookmarksDeletionManager.java
```diff
@@ -0,0 +1,234 @@
...
+
+  /**
+   * Clear state in case of redundancy (e.g., wipe).
+   */
+  public void clear() {
+    nonFolders.clear();
+    nonFolderCount = 0;
+    folders.clear();
+    folderCount = 0;
+    nonFolderParents.clear();
```
ncalexan: Why not `folderParents`?

src/main/java/org/mozilla/gecko/sync/repositories/delegates/RepositorySessionStoreDelegate.java
```diff
@@ -50,6 +50,9 @@
  */
 public interface RepositorySessionStoreDelegate {
   public void onRecordStoreFailed(Exception ex);
+
+  // Optionally called with an equivalent (but not necessarily identical) record
+  // when a store has succeeded.
```
ncalexan, line 55: I don't think you're even doing this.  As written, you can't assume that the argument to `onRecordStoreSucceeded` has _any_ data attached to it.  Actually, I take that back -- these records were deleted, so you can't assume they have data anyway.  Carry on, good sir.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksDataAccessor.java
```diff
@@ -119,7 +144,44 @@ protected void updateParentAndPosition(String guid, long newParentId, long posit
...
+    final int idIndex = c.getColumnIndexOrThrow(BrowserContract.Bookmarks._ID);
+    while (!c.isAfterLast()) {
+      out.put(c.getString(guidIndex), c.getLong(idIndex));
+      c.moveToNext();
+    }
+    return out;
+  }
+
+  /**
+   * Move all the children of the folders with IDs in `from` to have parent `to`.
```
ncalexan: nit: grammar.

ncalexan:
And have a random test.  I still want to see your concrete test, but the one I wrote definitely exercises the code, including checking that callbacks get called.

ncalexan:
Address nits and add more tests, then r=nalexander.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
rnewman: I've been holding out on this because I would like our store tracking to be timestamp aware — that is, when we track a record to ignore, we only ignore it if it doesn't get modified again at a later time, and that allows us to mostly dispense with untracking.

But yes, it's horrible, so perhaps I'll find a better way to fix that later.

src/main/java/org/mozilla/gecko/sync/repositories/android/BookmarksDeletionManager.java
rnewman, line 125: It will be!

PR #155, opened 2012-04-11
Bug 709385 - Passwords repository layer

liuche:
Fixed infinite loop problem, added more tests.


rnewman:
Rebased down, review comments self-addressed, and merged.

PR #158, opened 2012-04-12
Bug 742590 - TestClientsDataDelegate creates an account but doesn't clean up.

ncalexan:


rnewman:
If tests pass with this, then r+.

PR #159, opened 2012-04-12
Bug 744660 - bulk insert history extender

ncalexan:
Made bulkInsert call store (which update-and-inserts) in a single DB transaction.  It's not _really_ bulkInsert, but it is what we mean by bulkInsert.


rnewman:
13:46:07 @rnewman nalexander: I'd rather not rely on the return value of getCachedWritableDatabase
13:46:14 @rnewman which is called again within the transaction inside store
13:46:39 @rnewman let's lift out a separate method that takes a DB param

rnewman:
r+ after comment nit.

PR #160, opened 2012-04-12
Bug 739416 - Allow sending of commands.

emtwo:
Still needs some polishing but is mostly done.

### DIFF
--- a/src/main/java/org/mozilla/android/sync/test/helpers/CommandHelpers.java
+++ b/src/main/java/org/mozilla/android/sync/test/helpers/CommandHelpers.java
@@ -1,3 +1,6 @@
+/* Any copyright is dedicated to the Public Domain.
+   http://creativecommons.org/publicdomain/zero/1.0/ */
+
 package org.mozilla.android.sync.test.helpers;
 
 import org.json.simple.JSONArray;
--- a/src/main/java/org/mozilla/android/sync/test/helpers/CommandHelpers.java
+++ b/src/main/java/org/mozilla/android/sync/test/helpers/CommandHelpers.java
@@ -1,0 +1,37 @@
+package org.mozilla.android.sync.test.helpers;
+
+import org.json.simple.JSONArray;
+import org.mozilla.gecko.sync.CommandProcessor.Command;
+
+public class CommandHelpers {
+
+  @SuppressWarnings("unchecked")
+  public static Command getCommand1() {
+    JSONArray args = new JSONArray();
+    args.add("argsA");
+    return new Command("displayURI", args);
+  }
+
+  @SuppressWarnings("unchecked")
+  public static Command getCommand2() {
+    JSONArray args = new JSONArray();
+    args.add("argsB");
+    return new Command("displayURI", args);
+  }
+
+  @SuppressWarnings("unchecked")
+  public static Command getCommand3() {
+    JSONArray args = new JSONArray();
+    args.add("argsC");
+    return new Command("displayURI", args);
+  }
+
+  @SuppressWarnings("unchecked")
+  public static Command getCommand4() {
+    JSONArray args = new JSONArray();
+    args.add("URI of Page");
+    args.add("Sender ID");
+    args.add("Title of Page");
+    return new Command("displayURI", args);
+  }
+}
--- a/src/main/java/org/mozilla/gecko/sync/CommandProcessor.java
+++ b/src/main/java/org/mozilla/gecko/sync/CommandProcessor.java
@@ -179,29 +179,29 @@ public class CommandProcessor {
     }
   }
 
-  public void displayURI(List<String> args, Context context) {
-    // These two args are guaranteed to exist by trusting the client sender.
-    String uri = args.get(0);
-    String clientId = args.get(1);
+  public static void displayURI(final List<String> args, final Context context) {
+    // We trust the client sender that these exist.
+    final String uri = args.get(0);
+    final String clientId = args.get(1);
 
-    Logger.info(LOG_TAG, "Received a URI for display: " + uri + " from " + clientId);
+    Logger.pii(LOG_TAG, "Received a URI for display: " + uri + " from " + clientId);
 
     String title = null;
     if (args.size() == 3) {
       title = args.get(2);
     }
 
-    // Get NotificationManager.
-    String ns = Context.NOTIFICATION_SERVICE;
-    NotificationManager mNotificationManager = (NotificationManager)context.getSystemService(ns);
+    final String ns = Context.NOTIFICATION_SERVICE;
+    final NotificationManager notificationManager = (NotificationManager) context.getSystemService(ns);
 
-    // Create a Notficiation.
-    int icon = R.drawable.sync_ic_launcher;
+    // Create a Notificiation.
+    final int icon = R.drawable.sync_ic_launcher;
     String notificationTitle = context.getString(R.string.sync_new_tab);
     if (title != null) {
       notificationTitle = notificationTitle.concat(": " + title);
     }
-    long when = System.currentTimeMillis();
+
+    final long when = System.currentTimeMillis();
     Notification notification = new Notification(icon, notificationTitle, when);
     notification.flags = Notification.FLAG_AUTO_CANCEL;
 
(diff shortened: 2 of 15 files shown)


emtwo:
Now uploading all remote records together and waiting until they complete (last 2 commits)

src/main/java/org/mozilla/android/sync/test/helpers/CommandHelpers.java
```diff
@@ -0,0 +1,37 @@
+package org.mozilla.android.sync.test.helpers;
```
rnewman, line 4: License.

src/main/java/org/mozilla/gecko/sync/CommandProcessor.java
```diff
@@ -78,19 +96,62 @@ protected Command parseCommand(ExtendedJSONObject unparsedCommand) {
...
+    Logger.info(LOG_TAG, "In sendCommand.");
+    CommandRunner commandData = commands.get(command.commandType);
+
+    // Don't send commands that we don't know about.
+    if (commandData == null) {
+      Logger.error(LOG_TAG, "Unknown command to send: " + command);
+      return;
+    }
+    // Don't send a command with the wrong number of arguments.
+    else if (command.args == null || command.args.size() != commandData.argCount) {
```
rnewman: This would probably be better factored out in some way — that is, pre-validate the `Command` that's input, or have a `CommandRunner` able to validate arg counts.

src/main/java/org/mozilla/gecko/sync/repositories/android/ClientsDatabase.java
```diff
@@ -41,27 +52,55 @@ public ClientsDatabase(Context context) {
...
 
+  public void wipeDB() {
+    SQLiteDatabase db = this.getCachedWritableDatabase();
+    onUpgrade(db, SCHEMA_VERSION, SCHEMA_VERSION);
+  }
+
+  public void wipeClientsTable() {
+    SQLiteDatabase db = this.getCachedWritableDatabase();
+    db.execSQL("DROP TABLE IF EXISTS " + TBL_CLIENTS);
+    createClientsTable(db);
```
rnewman: What's wrong with `"DELETE FROM " + TBL_CLIENTS`?

src/main/java/org/mozilla/gecko/sync/repositories/android/ClientsDatabase.java
```diff
@@ -84,23 +123,82 @@ public void store(String profileId, ClientRecord record) {
...
+    try {
+      cur = this.fetchSpecificCommand(accountGUID, command, args);
+
+      if (cur != null && cur.moveToFirst()) {
+        Logger.debug(LOG_TAG, "Command already exists in database.");
+        return;
+      }
+    } finally {
+      if (cur != null) {
+        cur.close();
```
rnewman: This fragment can be rewritten to not require these checks.

src/main/java/org/mozilla/gecko/sync/repositories/android/ClientsDatabase.java
```diff
@@ -84,23 +123,82 @@ public void store(String profileId, ClientRecord record) {
...
+    cv.put(COL_COMMAND, command);
+    if (args == null) {
+      cv.put(COL_ARGS, "[]");
+    } else {
+      cv.put(COL_ARGS, args);
+    }
+
+    Cursor cur = null;
+    try {
+      cur = this.fetchSpecificCommand(accountGUID, command, args);
```
rnewman: This will never return null.

src/main/java/org/mozilla/gecko/sync/repositories/android/ClientsDatabaseAccessor.java
```diff
@@ -77,6 +85,50 @@ public ClientRecord fetch(String accountGUID) throws NullCursorException {
     }
   }
 
+  public List<Command> fetchAllCommands() throws NullCursorException {
+    List<Command> commands = new ArrayList<Command>();
+    Cursor cur = null;
+    try {
+      cur = db.fetchAllCommands();
```
rnewman: Similary.

src/main/java/org/mozilla/gecko/sync/repositories/android/ClientsDatabaseAccessor.java
```diff
@@ -77,6 +85,50 @@ public ClientRecord fetch(String accountGUID) throws NullCursorException {
...
+      if (cur == null || !cur.moveToFirst()) {
+        return Collections.unmodifiableList(commands);
+      }
+      while(!cur.isAfterLast()) {
```
rnewman: Nit: spacing.

src/main/java/org/mozilla/gecko/sync/repositories/android/ClientsDatabaseAccessor.java
```diff
@@ -77,6 +85,50 @@ public ClientRecord fetch(String accountGUID) throws NullCursorException {
...
+        cur.close();
+      }
+    }
+  }
+
+  public List<Command> fetchCommandsForClient(String accountGUID) throws NullCursorException {
+    List<Command> commands = new ArrayList<Command>();
+    Cursor cur = null;
+    try {
+      cur = db.fetchCommandsForClient(accountGUID);
```
rnewman: Similarly.

src/main/java/org/mozilla/gecko/sync/repositories/android/ClientsDatabaseAccessor.java
```diff
@@ -77,6 +85,50 @@ public ClientRecord fetch(String accountGUID) throws NullCursorException {
...
+      if (cur == null || !cur.moveToFirst()) {
+        return Collections.unmodifiableList(commands);
+      }
+      while(!cur.isAfterLast()) {
```
rnewman: Similarly.

src/main/java/org/mozilla/gecko/sync/stage/SyncClientsEngineStage.java
```diff
@@ -313,6 +349,57 @@ protected void processCommands(JSONArray commands) {
...
+      this.uploadClientRecord(cryptoRecord);
+      return;
+    }
+
+    JSONArray cryptoRecords = new JSONArray();
+    long latestLastModified = 0;
+    for (ClientRecord record : toUpload) {
+      Logger.trace(LOG_TAG, "Record " + record.guid + " is being uploaded" );
+      if (record.lastModified > latestLastModified) {
+        latestLastModified = record.lastModified;
```
rnewman: I don't think this is right. If I fetch at time 1000:
- record A: 110
- record B: 220
- record C: 330

and send commands to A and C, I'll set a header of 330, not 1000, and the request will fail.

You want to save the timestamp from when you got these records via HTTP GET, and use that — it'll guarantee that nobody will have touched the server since you fetched them.

emtwo: Ah, I think what you're implying here is that the overall timestamp of a collection is not the same as the timestamp of the last modified record?

performing an HTTP GET will actually update the collection's overall last modified timestamp but no changes are made to the individual records' timestamps? (wish server documentation was more specific...)

rnewman: > Ah, I think what you're implying here is that the overall timestamp of a collection is not the same as the timestamp of the last modified record?

Yes. If you pass X-If-Unmodified-Since on a POST, the server will check it against the collection timestamp, so the value you pass in has to be the timestamp you got when you fetched these records. If you pass in the timestamp from one of the individual records, it will always be earlier than the server's modified time, unless it was the _last_ uploaded record.

That is, your code will work if you upload _all_ client records, or include the most recently modified one in the set, but it won't work in the general case.

> performing an HTTP GET will actually update the collection's overall last modified timestamp but no changes are made to the individual records' timestamps? (wish server documentation was more specific...)

GET won't change any timestamps. What you're trying to do is detect when someone else might have modified one of these records since you fetched it. The mechanism to do that when you're processing multiple records is to track the time at which you _know_ nobody modified it.

For an individual record you can use its modified time directly.

emtwo: Excellent! Thank you for the thorough explanation! :)

emtwo:
Reviewed your commits & they look good. Added the timestamp change.

emtwo:
Reverted sendCommand changes & added a comment.

rnewman:
Ugh. This is really hairy, and will take a long time to un-bitrot.

rnewman:
Obsoleted by #211.

PR #161, opened 2012-04-12
Bug 740193 - Send tab to device: push tabs to other devices

emtwo:
Also still needs some more polish but is mostly good.


emtwo:
rebased & updated so there are no longer problems with sending a command after an upgrade or restart.

PR #162, opened 2012-04-12
Bug 743098 - migration before bookmarks and history

ncalexan:
Due to Fennec passwords CP crashiness, hasn't been tested on device.

This disables the engine if the Fennec dbs aren't marked as migrated.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/repositories/android/FennecControlHelper.java
+++ b/src/main/java/org/mozilla/gecko/sync/repositories/android/FennecControlHelper.java
@@ -1,0 +1,112 @@
+package org.mozilla.gecko.sync.repositories.android;
+
+import org.mozilla.gecko.db.BrowserContract.Control;
+import org.mozilla.gecko.sync.Logger;
+import org.mozilla.gecko.sync.repositories.NoContentProviderException;
+
+import android.content.ContentProviderClient;
+import android.content.Context;
+import android.database.Cursor;
+import android.net.Uri;
+
+/**
+ * This class provides an interface to Fenec's control content provider, which
+ * exposes some of Fennec's internal state.
+ */
+public class FennecControlHelper {
+  public static String LOG_TAG = "FennecControlHelper";
+
+  protected static Uri FENNEC_CONTROL_URI = Control.CONTENT_URI;
+
+  protected ContentProviderClient providerClient;
+  protected final RepoUtils.QueryHelper queryHelper;
+
+  public FennecControlHelper(Context context)
+      throws NoContentProviderException {
+    providerClient = acquireContentProvider(context);
+    queryHelper = new RepoUtils.QueryHelper(context, FENNEC_CONTROL_URI, LOG_TAG);
+  }
+
+  /**
+   * Acquire the content provider client.
+   * <p>
+   * The caller is responsible for releasing the client.
+   *
+   * @param context The application context.
+   * @return The <code>ContentProviderClient</code>.
+   * @throws NoContentProviderException
+   */
+  public static ContentProviderClient acquireContentProvider(final Context context)
+      throws NoContentProviderException {
+    Uri uri = FENNEC_CONTROL_URI;
+    ContentProviderClient client = context.getContentResolver().acquireContentProviderClient(uri);
+    if (client == null) {
+      throw new NoContentProviderException(uri);
+    }
+    return client;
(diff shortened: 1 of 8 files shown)


src/main/java/org/mozilla/gecko/sync/repositories/android/FennecControlHelper.java
```diff
@@ -0,0 +1,92 @@
+package org.mozilla.gecko.sync.repositories.android;
```
rnewman, line 5: License.

src/main/java/org/mozilla/gecko/sync/repositories/android/FennecControlHelper.java
```diff
@@ -0,0 +1,92 @@
...
+    Cursor cursor = null;
+    try {
+      cursor = queryHelper.safeQuery(providerClient, ".isColumnMigrated(" + column + ")", columns, null, null, null);
+      if (!cursor.moveToFirst()) {
+        return false;
+      }
+      int value = RepoUtils.getIntFromCursor(cursor, column);
+      return value > 0;
+    } finally {
+      if (cursor != null) {
```
rnewman: If you move the `cursor = queryHelper…` outside the `try`, you can rely on the behavior of `safeQuery` and your `throws` clause and get rid of this condition.

src/main/java/org/mozilla/gecko/sync/stage/AndroidBrowserBookmarksServerSyncStage.java
```diff
@@ -80,4 +84,31 @@ protected Repository getLocalRepository() {
...
+
+  protected boolean areBookmarksMigrated() {
+    // Avoid colliding with a migrating Fennec.
+    if (session == null || session.getContext() == null) {
+      return false;
+    }
+    FennecControlHelper control = null;
+    boolean migrated = false;
+    try {
+      control = new FennecControlHelper(session.getContext());
```
rnewman: Lift this out of the `try`.

src/main/java/org/mozilla/gecko/sync/stage/AndroidBrowserHistoryServerSyncStage.java
```diff
@@ -80,4 +84,31 @@ protected Repository getRemoteRepository() throws URISyntaxException {
...
+  protected boolean isHistoryMigrated() {
+    // Avoid colliding with a migrating Fennec.
+    if (session == null || session.getContext() == null) {
+      return false;
+    }
+    FennecControlHelper control = null;
+    boolean migrated = false;
+    try {
+      control = new FennecControlHelper(session.getContext());
+      migrated = control.isHistoryMigrated();
```
rnewman: So why not embed all of this safety in `isFooMigrated()`, and add a static version, so you can just do:

```
boolean migrated = FennecControlHelper.isHistoryMigrated(session.getContext());
if (!migrated) {
  // Log
}
return migrated;
```

where

```
public static boolean isHistoryMigrated(Context context) {
  FennecControlHelper h = new FennecControlHelper(context);
  try {
    …
  } finally {
    h.releaseProviders();
  }
}
```

It's the same between both content providers, and you just release…

test/src/org/mozilla/gecko/sync/repositories/android/test/TestFennecControlHelper.java
```diff
@@ -0,0 +1,20 @@
...
+
+public class TestFennecControlHelper extends AndroidSyncTestCase {
+  public void testIsMigrated() throws NoContentProviderException, NullCursorException, RemoteException {
+    FennecControlHelper control = new FennecControlHelper(getApplicationContext());
+    assertTrue(control.areBookmarksMigrated());
+    // Might take a few shots, but we should be able to migrate if we haven't already.
+    int i = 15;
+    while (!control.isHistoryMigrated()) {
+      assertTrue(i-- > 0);
+    }
```
rnewman: Print some log output before this warning that this test will fail if migration takes time!

ncalexan:
I implemented static access very slightly differently than you suggested, but I think it gets the same idea.  We could remove the instance altogether; I was looking ahead to being able to query more things (especially profile data) and assumed we would want to ask for lots of things in a single sync, so a stateful object seemed right.  Let me know if we'd rather go static.

PR #163, opened 2012-04-13
Bug 742104 - turn on forms

ncalexan:
- cleans forms batching to get rid of an unneeded runnable.
- fetches at most 5000 form history items (too high?) and applies in batches of 200.
- INCLUDES COMMIT TO DISABLE PASSWORDS.

This is very snappy on my Transformer Prime now.


rnewman:
This is done and merged, with the exception of disabling passwords, which I'm doing in pure-hg.

PR #164, opened 2012-04-17
Bug 709323 - key changes

ncalexan:
Calls resetClient() with appropriate engines on key changes.

_Warning:_ this has not had any device testing; would love direction on how much effort to put into unit/integration testing this.  I can definitely factor this into a more testable form, but we won't end up testing much actual code after all the stubbing is done.


ncalexan:
Re: testing: https://bugzilla.mozilla.org/show_bug.cgi?id=709321 (generate and upload keys) will be closed by writing tests, so I'll unit test the whole damn stage and close both of these.

ncalexan:
Closed in favour of https://github.com/mozilla-services/android-sync/pull/166

PR #166, opened 2012-04-17
Bugs 709321 and 709323 - keys

ncalexan:
This actually fixed a bug in the existing upload code -- we encrypted our upload record twice.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/GlobalSession.java
+++ b/src/main/java/org/mozilla/gecko/sync/GlobalSession.java
@@ -403,10 +403,6 @@ public class GlobalSession implements CredentialsSource, PrefsSource, HttpRespon
     }
   }
 
-  public InfoCollections getInfoCollections() {
-    return this.config.infoCollections;
-  }
-
   public void fetchInfoCollections(InfoCollectionsDelegate callback) throws URISyntaxException {
     if (this.config.infoCollections == null) {
       this.config.infoCollections = new InfoCollections(config.infoURL(), credentials());
--- a/src/main/java/org/mozilla/gecko/sync/GlobalSession.java
+++ b/src/main/java/org/mozilla/gecko/sync/GlobalSession.java
@@ -403,6 +403,10 @@ public class GlobalSession implements CredentialsSource, PrefsSource, HttpRespon
     }
   }
 
+  public InfoCollections getInfoCollections() {
+    return this.config.infoCollections;
+  }
+
   public void fetchInfoCollections(InfoCollectionsDelegate callback) throws URISyntaxException {
     if (this.config.infoCollections == null) {
       this.config.infoCollections = new InfoCollections(config.infoURL(), credentials());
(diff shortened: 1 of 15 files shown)


rnewman:
Never been tested until now :D

rnewman:
Looks pretty good.

src/main/java/org/mozilla/gecko/sync/GlobalSession.java
```diff
@@ -563,9 +578,9 @@ public void handleSuccess(MetaGlobal global, SyncStorageResponse response) {
 
             // Generate and upload new keys.
             try {
-              session.uploadKeys(CollectionKeys.generateCollectionKeys().asCryptoRecord(), new KeyUploadDelegate() {
+              session.uploadKeys(CollectionKeys.generateCollectionKeys(), new KeyUploadDelegate() {
                 @Override
-                public void onKeysUploaded() {
+                public void onKeysUploaded(CollectionKeys keys, long timestamp) {
                   // Now we can download them.
```
rnewman: See this?

src/main/java/org/mozilla/gecko/sync/stage/EnsureCrypto5KeysStage.java
```diff
@@ -161,10 +221,13 @@ public void handleRequestError(Exception ex) {
...
 
   @Override
-  public void onKeysUploaded() {
-    Logger.debug(LOG_TAG, "New keys uploaded. Starting stage again to fetch them.");
+  public void onKeysUploaded(CollectionKeys keys, long timestamp) {
+    Logger.debug(LOG_TAG, "New keys uploaded. Persisting before starting stage again.");
     try {
       retrying = true;
+      PersistedCrypto5Keys pck = session.config.persistedCryptoKeys();
+      pck.persistKeys(keys);
```
rnewman: That means we shouldn't do this.

The way we generate keys is:
- Observe that the server has no keys
- Generate and upload some keys, but don't store them locally
- Download keys from the server.

This way we're totally protected from race conditions: we _never_ end up with keys locally that we didn't download from the server. Y'see?

rnewman: This is particularly important in light of the comment elsewhere:

```
        // It would be good to set the X-If-Unmodified-Since header to `timestamp`
        // for this PUT to ensure at least some level of transactionality.
        // Unfortunately, the servers don't support it after a wipe right now
        // (bug 693893), so we're going to defer this until bug 692700.
```

We could race generating and uploading keys, and we'd be convinced that we just uploaded some. Leave them non-persisted. `onFreshStart` will restart the session and redownload anyway.

ncalexan:
Re-based and addressed review comments.  I am happy to factor this down into the original 4 or 5 commits after approval.

ncalexan:
Worth noting: this has not been device tested, but it will be when I get enough of the clean server stuff implemented to test the lot.

rnewman:
Beautifully rebased into two separate sets of commits, if I do say so myself, and merged.

PR #168, opened 2012-04-19
Bug 730636 - Implement resetting

rnewman:
No harm in having a pull request here for this.


ncalexan:
r=nalexander for this.

rnewman:
Changed since last pull request:

Resetting and wiping can require access to the `GlobalSession`. For example, in `SyncClientsEngineStage`:

```
  public void resetLocal() {
    // Clear timestamps and local data.
    session.config.persistServerClientRecordTimestamp(0L);
    session.getClientsDelegate().setClientsCount(0);
  …
```

Previously, the session instance was passed in `execute`. If one were to call `resetLocal` prior to executing the stage — which we would, as we'd be handling the reset command earlier — the session would be null and this wouldn't work.

Commits b6ae4b3..da7a781 reflect this alteration. I separated out da7a781 because it's not a trivial transformation. Stages now have a `final` reference to their session, and use this appropriately.

Similarly, the clients stage data accessor was lifted out.

Subsequently I've added a test for wiping the clients stage, because that logic should be defined and exercised somewhere. This is a separate integration test, not included in the clients stage unit tests.

rnewman:
Note that the non-clients stages (`ServerSyncStage`) implement wiping and resetting by instantiating a `RepositorySession` and manipulating the `SynchronizerConfiguration` respectively, and thus also need their session.

PR #170, opened 2012-04-19
Bug 713542 - batch bookmark inserts and Bug 743153 - Use ContentUris.parseId instead of RepoUtils.getAndroidIdFromUri.

ncalexan:
I have not explicitly unit tested that this preserves positions (beyond what was already tested in BookmarkPositioningTest).

I have tested this on device, and batch inserting of bookmarks and folders worked well, and positioning worked as well.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/Utils.java
+++ b/src/main/java/org/mozilla/gecko/sync/Utils.java
@@ -1,40 +1,6 @@
-/* ***** BEGIN LICENSE BLOCK *****
- * Version: MPL 1.1/GPL 2.0/LGPL 2.1
- *
- * The contents of this file are subject to the Mozilla Public License Version
- * 1.1 (the "License"); you may not use this file except in compliance with
- * the License. You may obtain a copy of the License at
- * http://www.mozilla.org/MPL/
- *
- * Software distributed under the License is distributed on an "AS IS" basis,
- * WITHOUT WARRANTY OF ANY KIND, either express or implied. See the License
- * for the specific language governing rights and limitations under the
- * License.
- *
- * The Original Code is Android Sync Client.
- *
- * The Initial Developer of the Original Code is
- * the Mozilla Foundation.
- * Portions created by the Initial Developer are Copyright (C) 2011
- * the Initial Developer. All Rights Reserved.
- *
- * Contributor(s):
- * Jason Voll <jvoll@mozilla.com>
- * Richard Newman <rnewman@mozilla.com>
- *
- * Alternatively, the contents of this file may be used under the terms of
- * either the GNU General Public License Version 2 or later (the "GPL"), or
- * the GNU Lesser General Public License Version 2.1 or later (the "LGPL"),
- * in which case the provisions of the GPL or the LGPL are applicable instead
- * of those above. If you wish to allow use of your version of this file only
- * under the terms of either the GPL or the LGPL, and not to allow others to
- * use your version of this file under the terms of the MPL, indicate your
- * decision by deleting the provisions above and replace them with the notice
- * and other provisions required by the GPL or the LGPL. If you do not delete
- * the provisions above, a recipient may use your version of this file under
- * the terms of any one of the MPL, the GPL or the LGPL.
- *
- * ***** END LICENSE BLOCK ***** */
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 package org.mozilla.gecko.sync;
 
@@ -308,5 +274,19 @@ public class Utils {
       }
     }
     return out;
+  }
+
+  // Because TextUtils.join is not stubbed.
+  public static String join(String delimiter, String... strings) {
+    StringBuilder sb = new StringBuilder();
+    int i = 0;
+    for (String string : strings) {
+      sb.append(string);
+      i += 1;
+      if (i < strings.length) {
+        sb.append(delimiter);
+      }
+    }
+    return sb.toString();
   }
 }
--- a/src/main/java/org/mozilla/gecko/sync/Utils.java
+++ b/src/main/java/org/mozilla/gecko/sync/Utils.java
@@ -12,6 +12,7 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.security.SecureRandom;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
@@ -277,16 +278,24 @@ public class Utils {
   }
 
   // Because TextUtils.join is not stubbed.
-  public static String join(String delimiter, String... strings) {
+  public static String toDelimitedString(String delimiter, Collection<String> items) {
+    if (items == null || items.size() == 0) {
+      return "";
+    }
+
     StringBuilder sb = new StringBuilder();
     int i = 0;
-    for (String string : strings) {
+    int c = items.size();
+    for (String string : items) {
       sb.append(string);
-      i += 1;
-      if (i < strings.length) {
+      if (++i < c) {
         sb.append(delimiter);
       }
     }
     return sb.toString();
+  }
+
+  public static String toCommaSeparatedString(Collection<String> items) {
+    return toDelimitedString(", ", items);
   }
 }
--- a/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
+++ b/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
@@ -820,9 +820,10 @@ public class AndroidBrowserBookmarksRepositorySession extends AndroidBrowserRepo
         Logger.warn(LOG_TAG, "Enqueued but didn't insert the following guids: " +
             "(" + Utils.toCommaSeparatedString(eSet) + ")");
       }
+
       iSet.removeAll(new HashSet<String>(enqueuedGuids));
       if (iSet.isEmpty()) {
-        Logger.debug(LOG_TAG, "All inserted guids were enqueued.");
+        // Should always be the case.
       } else {
         Logger.warn(LOG_TAG, "Inserted but didn't enqueue the following guids: " +
             "(" + Utils.toCommaSeparatedString(iSet) + ")");
--- a/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
+++ b/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
@@ -7,6 +7,7 @@ package org.mozilla.gecko.sync.repositories.android;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -723,7 +724,9 @@ public class AndroidBrowserBookmarksRepositorySession extends AndroidBrowserRepo
     }
   }
 
-  public class AndroidBrowserBookmarksInsertionManager extends BookmarksInsertionManager{
+  protected final ArrayList<String> insertedGuids = new ArrayList<String>();
+
+  public class AndroidBrowserBookmarksInsertionManager extends BookmarksInsertionManager {
     public AndroidBrowserBookmarksInsertionManager(int flushThreshold, Set<String> writtenFolders) {
       super(flushThreshold, writtenFolders);
     }
@@ -740,6 +743,7 @@ public class AndroidBrowserBookmarksRepositorySession extends AndroidBrowserRepo
       updateBookkeeping(toStore);
       trackRecord(toStore);
       delegate.onRecordStoreSucceeded(toStore);
+      insertedGuids.add(record.guid);
     }
 
     @Override
@@ -753,16 +757,20 @@ public class AndroidBrowserBookmarksRepositorySession extends AndroidBrowserRepo
         trackRecord(toStore);
         delegate.onRecordStoreSucceeded(toStore);
         toStores.add(toStore);
+        insertedGuids.add(record.guid);
       }
 
       dataAccessor.bulkInsert(toStores);
     }
   }
 
+  protected ArrayList<String> enqueuedGuids = new ArrayList<String>();
+
   @Override
   protected void insert(Record record) throws NoGuidForIdException, NullCursorException, ParentNotFoundException {
     try {
       insertionManager.enqueueRecord((BookmarkRecord) record);
+      enqueuedGuids.add(record.guid);
     } catch (Exception e) {
       throw new NullCursorException(e);
     }
@@ -783,12 +791,35 @@ public class AndroidBrowserBookmarksRepositorySession extends AndroidBrowserRepo
 
   protected void flushQueues() {
     long now = now();
-    Logger.debug(LOG_TAG, "Applying insertions.");
+    Logger.debug(LOG_TAG, "Applying remaining insertions.");
     try {
       insertionManager.flushAll(now);
-      Logger.debug(LOG_TAG, "Done applying insertions.");
+      Logger.debug(LOG_TAG, "Done applying remaining insertions.");
     } catch (Exception e) {
-      Logger.error(LOG_TAG, "Unable to apply insertions.", e);
+      Logger.warn(LOG_TAG, "Unable to apply remaining insertions.", e);
+    }
+
+    if (enqueuedGuids.size() != insertedGuids.size()) {
+      Logger.warn(LOG_TAG, "Enqueued " + enqueuedGuids.size() + " records, but inserted " +
+          insertedGuids.size() + " records.");
+      Set<String> eSet = new HashSet<String>(enqueuedGuids);
+      Set<String> iSet = new HashSet<String>(insertedGuids);
+      eSet.removeAll(iSet);
+      if (eSet.isEmpty()) {
+        Logger.debug(LOG_TAG, "All enqueued guids were inserted.");
+      } else {
+        Logger.warn(LOG_TAG, "Enqueued but didn't insert the following guids: (" +
+            Utils.join(", ", eSet.toArray(new String[eSet.size()])) + ")");
+      }
+      iSet.removeAll(new HashSet<String>(enqueuedGuids));
+      if (iSet.isEmpty()) {
+        Logger.debug(LOG_TAG, "All inserted guids were enqueued.");
+      } else {
+        Logger.warn(LOG_TAG, "Inserted but didn't enqueue the following guids: (" +
+            Utils.join(", ", iSet.toArray(new String[eSet.size()])) + ")");
+      }
+    } else {
+      Logger.debug(LOG_TAG, "Enqueued and inserted " + enqueuedGuids.size() + " records.");
     }
 
     Logger.debug(LOG_TAG, "Applying deletions.");
--- a/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
+++ b/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
@@ -607,8 +607,7 @@ public class AndroidBrowserBookmarksRepositorySession extends AndroidBrowserRepo
     // TODO: persist records that fail to insert for later retry.
     ArrayList<Record> toStores = new ArrayList<Record>(records.size());
     for (Record record : records) {
-      Record toStore = (BookmarkRecord) prepareRecord(record);
-      toStores.add(toStore);
+      toStores.add(prepareRecord(record));
     }
 
     try {
--- a/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
+++ b/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
@@ -574,15 +574,15 @@ public class AndroidBrowserBookmarksRepositorySession extends AndroidBrowserRepo
     // A folder that is *not* deleted needs its androidID updated, so that
     // updateBookkeeping can re-parent, etc.
     Record toStore = prepareRecord(record);
-    Uri recordURI = dbHelper.insert(toStore);
-    if (recordURI == null) {
-      delegate.onRecordStoreFailed(new RuntimeException("Got null URI inserting folder with guid " + toStore.guid + "."));
-      return false;
-    }
-    toStore.androidID = ContentUris.parseId(recordURI);
-    Logger.debug(LOG_TAG, "Inserted folder with guid " + toStore.guid + " as androidID " + toStore.androidID);
-
     try {
+      Uri recordURI = dbHelper.insert(toStore);
+      if (recordURI == null) {
+        delegate.onRecordStoreFailed(new RuntimeException("Got null URI inserting folder with guid " + toStore.guid + "."));
+        return false;
+      }
+      toStore.androidID = ContentUris.parseId(recordURI);
+      Logger.debug(LOG_TAG, "Inserted folder with guid " + toStore.guid + " as androidID " + toStore.androidID);
+
       updateBookkeeping(toStore);
     } catch (Exception e) {
       delegate.onRecordStoreFailed(e);
--- a/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
+++ b/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
@@ -926,8 +926,7 @@ public class AndroidBrowserBookmarksRepositorySession extends AndroidBrowserRepo
       try {
         // Clear our queued deletions.
         deletionManager.clear();
-        // Clearing the insertion manager is not supported yet. To clear, you need
-        // to reset the set of written folder GUIDs, which will need to be re-computed.
+        insertionManager.clear();
         super.run();
       } catch (Exception ex) {
         delegate.onWipeFailed(ex);
--- a/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
+++ b/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
@@ -7,6 +7,7 @@ package org.mozilla.gecko.sync.repositories.android;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.TreeMap;
@@ -35,6 +36,8 @@ import android.database.Cursor;
 public class AndroidBrowserBookmarksRepositorySession extends AndroidBrowserRepositorySession {
 
   public static final int DEFAULT_DELETION_FLUSH_THRESHOLD = 50;
+  public static final int DEFAULT_INSERTION_FLUSH_THRESHOLD = 50;
+
   // TODO: synchronization for these.
   private HashMap<String, Long> parentGuidToIDMap = new HashMap<String, Long>();
   private HashMap<Long, String> parentIDToGuidMap = new HashMap<Long, String>();
@@ -101,6 +104,7 @@ public class AndroidBrowserBookmarksRepositorySession extends AndroidBrowserRepo
   private AndroidBrowserBookmarksDataAccessor dataAccessor;
 
   protected BookmarksDeletionManager deletionManager;
+  protected BookmarksInsertionManager insertionManager;
 
   /**
    * An array of known-special GUIDs.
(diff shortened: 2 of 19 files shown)


ncalexan:
Hmm, something is not working quite right.  Hold off on this until I get a chance to device test further.

ncalexan:
I have device tested this and run lots of random integration tests against it, and I think it works, but this is a high risk patch.

src/main/java/org/mozilla/gecko/sync/Utils.java
```diff
@@ -304,4 +270,18 @@ public static boolean sameArrays(JSONArray a, JSONArray b) {
     }
     return out;
   }
+
+  // Because TextUtils.join is not stubbed.
+  public static String join(String delimiter, String... strings) {
```
rnewman: This is called with a Collection or an array _everywhere_. Let's avoid variadics and the redundant allocation of arrays.

It's also only ever called with a comma for `delimiter`, so I suggest instead:

```
public static String toDelimitedString(String delimiter, Collection<String> items) {
  if (items == null || items.size() == 0) {
    return "";
  }

  StringBuilder sb = new StringBuilder();
  int i = 0;
  int c = items.size();
  for (String string : items) {
    sb.append(string);
    if (++i < c) {
      sb.append(delimiter);
    }
  }
  return sb.toString();
}

public static String toCommaSeparatedString(Collection<String> items) {
  return toDelimitedString(", ", items);
}
```

and then using `toCommaSeparatedString(coll)`.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
```diff
@@ -549,14 +557,18 @@ public void begin(RepositorySessionBeginDelegate delegate) throws InvalidSession
       cur.close();
     }
     deletionManager = new BookmarksDeletionManager(dataAccessor, DEFAULT_DELETION_FLUSH_THRESHOLD);
+    insertionManager = new AndroidBrowserBookmarksInsertionManager(DEFAULT_INSERTION_FLUSH_THRESHOLD,
+        parentGuidToIDMap.keySet());
```
rnewman: Add a comment that this `keySet` is backed by the `Map`, so it will reflect the current state.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
```diff
@@ -712,6 +724,58 @@ protected void updateBookkeeping(Record record) throws NoGuidForIdException,
...
+      super(flushThreshold, writtenFolders);
+    }
+
+    @Override
+    protected void insertFolder(BookmarkRecord record) throws Exception {
+      // A folder that is *not* deleted needs its androidID updated, so that
+      // updateBookkeeping can re-parent, etc.
+      Record toStore = prepareRecord(record);
+      Uri recordURI = dbHelper.insert(toStore);
+      toStore.androidID = ContentUris.parseId(recordURI);
```
rnewman: You should check for null here before passing to `parseId`. The result can be null.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
```diff
@@ -712,6 +724,58 @@ protected void updateBookkeeping(Record record) throws NoGuidForIdException,
...
+      updateBookkeeping(toStore);
+      trackRecord(toStore);
+      delegate.onRecordStoreSucceeded(toStore);
+      insertedGuids.add(record.guid);
+    }
+
+    @Override
+    protected void bulkInsertNonFolders(List<BookmarkRecord> records) throws Exception {
+      // All of these records are *not* deleted and *not* folders, so we don't
+      // need to update androidID at all!
```
rnewman: But we will need to track failures eventually…

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
```diff
@@ -712,6 +724,58 @@ protected void updateBookkeeping(Record record) throws NoGuidForIdException,
...
+      for (Record record : records) {
+        Record toStore = (BookmarkRecord) prepareRecord(record);
+        updateBookkeeping(toStore);
+        trackRecord(toStore);
+        delegate.onRecordStoreSucceeded(toStore);
+        toStores.add(toStore);
+        insertedGuids.add(record.guid);
+      }
+
+      dataAccessor.bulkInsert(toStores);
```
rnewman: You should do this before you do all of the post-insert work.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
```diff
@@ -725,10 +789,41 @@ protected void storeRecordDeletion(final Record record, final Record existingRec
...
+    try {
+      insertionManager.flushAll(now);
+      Logger.debug(LOG_TAG, "Done applying remaining insertions.");
+    } catch (Exception e) {
+      Logger.warn(LOG_TAG, "Unable to apply remaining insertions.", e);
+    }
+
+    if (enqueuedGuids.size() != insertedGuids.size()) {
+      Logger.warn(LOG_TAG, "Enqueued " + enqueuedGuids.size() + " records, but inserted " +
+          insertedGuids.size() + " records.");
```
rnewman: In what situation will this occur?

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
```diff
@@ -712,6 +724,58 @@ protected void updateBookkeeping(Record record) throws NoGuidForIdException,
...
+      trackRecord(toStore);
+      delegate.onRecordStoreSucceeded(toStore);
+      insertedGuids.add(record.guid);
+    }
+
+    @Override
+    protected void bulkInsertNonFolders(List<BookmarkRecord> records) throws Exception {
+      // All of these records are *not* deleted and *not* folders, so we don't
+      // need to update androidID at all!
+      ArrayList<Record> toStores = new ArrayList<Record>();
```
rnewman: Might as well do `new ArrayList<Record>(records)`, then mutate in place, to save resizing the array.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
```diff
@@ -794,6 +889,8 @@ public void run() {
       try {
         // Clear our queued deletions.
         deletionManager.clear();
+        // Clearing the insertion manager is not supported yet. To clear, you need
+        // to reset the set of written folder GUIDs, which will need to be re-computed.
```
rnewman: This is in `wipe`; we're blowing away all the data. That means we can just discard any pending insertions.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserHistoryRepositorySession.java
```diff
@@ -148,17 +148,16 @@ public void finish(final RepositorySessionFinishDelegate delegate) throws Inacti
...
-   *         records are flushed to the database.
    */
   @Override
-  protected Record insert(Record record) throws NoGuidForIdException, NullCursorException, ParentNotFoundException {
+  protected void insert(Record record) throws NoGuidForIdException, NullCursorException, ParentNotFoundException {
     HistoryRecord toStore = (HistoryRecord) prepareRecord(record);
     toStore.androidID = -111; // Hopefully this special value will make it easy to catch future errors.
     updateBookkeeping(toStore); // Does not use androidID -- just GUID -> String map.
+    trackRecord(toStore);
+    delegate.onRecordStoreSucceeded(toStore); // Technically this is too early.
```
rnewman: Can we make it not too early? :P

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserRepositoryDataAccessor.java
```diff
@@ -177,4 +179,38 @@ public void updateByGuid(String guid, ContentValues cv) {
...
+
+    int size = records.size();
+    ContentValues[] cvs = new ContentValues[size];
+    int index = 0;
+    for (Record record : records) {
+      cvs[index] = getContentValues(record);
+      index += 1;
+    }
+
+    // First update the history records.
```
rnewman: This comment doesn't look right.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserRepositoryDataAccessor.java
```diff
@@ -177,4 +179,38 @@ public void updateByGuid(String guid, ContentValues cv) {
...
+    int inserted = context.getContentResolver().bulkInsert(getUri(), cvs);
+    if (inserted == size) {
```
rnewman: You should return `inserted`.

src/main/java/org/mozilla/gecko/sync/repositories/android/BookmarksInsertionManager.java
```diff
@@ -0,0 +1,229 @@
...
+  public BookmarksInsertionManager(int flushThreshold, Set<String> writtenFolders) {
+    this.flushThreshold = flushThreshold;
+    this.writtenFolders.addAll(writtenFolders);
+  }
+
+  protected void addRecordWithUnwrittenParent(BookmarkRecord record) {
+    if (!waitingForParent.containsKey(record.parentID)) {
+      waitingForParent.put(record.parentID, new ArrayList<BookmarkRecord>());
+    }
+    waitingForParent.get(record.parentID).add(record);
```
rnewman: You can save work here:

```
ArrayList<BookmarkRecord> destination = waitingForParent.get(record.parentID);
if (destination == null) {
  destination = new ArrayList<BookmarkRecord>();
  waitingForParent.put(record.parentID, destination);
}
destination.add(record);
```

_but_ -- note that you should be using a `Map`, not an `ArrayList`, to implement set-like semantics (because we can't use the database to reconcile colliding inserts), or at the very least have a big comment explaining that we're never expecting two records with the same ID to appear in the same `RepositorySession` store sequence.

src/main/java/org/mozilla/gecko/sync/repositories/android/BookmarksInsertionManager.java
```diff
@@ -0,0 +1,229 @@
...
+  }
+
+  protected void recursivelyAddRecordAndChildren(BookmarkRecord record) {
+    Logger.debug(LOG_TAG, "Record has known parent with guid " + record.parentID + "; adding to insertion queue.");
+    readyToWrite.add(record);
```
rnewman: Note that you're relying on the order of `readyToWrite` applying to the bulk insert. That's fine, but best to document it.

src/main/java/org/mozilla/gecko/sync/repositories/android/BookmarksInsertionManager.java
```diff
@@ -0,0 +1,229 @@
...
+
+    if (record.isFolder()) {
+      writtenFolders.add(record.guid);
```
rnewman: And this is a slight lie. If the folder insertion fails, subsequent children won't be able to be inserted.

I perhaps suggest doing things a little differently: flush folders _immediately_, or batch together folders and flush them when a non-folder is encountered. We want to know if a folder insert fails before we try to do anything with the children.

src/main/java/org/mozilla/gecko/sync/repositories/android/BookmarksInsertionManager.java
```diff
@@ -0,0 +1,229 @@
...
+    // Write folders first.
+    ArrayList<BookmarkRecord> nonFolders = new ArrayList<BookmarkRecord>();
+    for (BookmarkRecord record : readyToWrite) {
+      if (!record.isFolder()) {
+        nonFolders.add(record);
+        continue;
+      }
+      // Folders are inserted one at a time.
+      try {
+        insertFolder(record);
```
rnewman: Again, if folders are inserted one at a time, I think we should bring them as early in the process as possible, and do it directly by calling `insertFolder`, rather than `enqueueRecord > enqueueFolder > incrementalFlush > flushReadyToWrite` etc.

Just call `insertFolder` at the top of `recursivelyAddRecordAndChildren`, which will avoid the lies in that method!

My suggestion might change if we actually did bulk insert folders, and could do so safely, but given that that's not the case, I'm inclined to have `readyToWrite` only ever contain non-folders, which will simplify this method significantly.

src/main/java/org/mozilla/gecko/sync/repositories/android/PasswordsRepositorySession.java
```diff
@@ -445,7 +446,7 @@ public PasswordRecord insert(PasswordRecord record) throws RemoteException {
     // record.timeLastUsed = now();
     ContentValues cv = getContentValues(record);
     Uri insertedUri = passwordsProvider.insert(BrowserContractHelpers.PASSWORDS_CONTENT_URI, cv);
-    record.androidID = RepoUtils.getAndroidIdFromUri(insertedUri);
+    record.androidID = ContentUris.parseId(insertedUri);
```
rnewman: Again, null check.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
ncalexan: So?  If it fails to insert, by definition it does not have a valid androidID.  And why would you need a stale androidID when tracking?

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
ncalexan: If there is an error in my logic, it is possible that we enqueue records that never get inserted (written to the db).  This will at least warn.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
ncalexan: Indeed.  The problem is that after the wipe, we need to re-calculate the set of folders written in the DB (which should be just places, unfiled, etc).  At the moment, this calculation is only done in `begin`.  I think it's legal to call `wipe` and then insert some things in the same session (which may already fail, depending on whether wipe preserves the special guids).  In this case, `parentGuidsToIDsMap` is out of date, as is the `insertionManager`'s internal map.  So I'm explicitly not handling this case yet.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
```diff
@@ -549,14 +557,24 @@ public void begin(RepositorySessionBeginDelegate delegate) throws InvalidSession
       cur.close();
     }
     deletionManager = new BookmarksDeletionManager(dataAccessor, DEFAULT_DELETION_FLUSH_THRESHOLD);
+
+    // We just crawled the database enumerating all folders; we'll start the
+    // insertion manager with exactly these folders as the known parents. From
+    // here on, there is no reference to <code>parentGuidToIDMap</code> in the
+    // insertion manager.
```
rnewman: This is incorrect: the insertion manager updates that map via methods in its enclosing instance. (Try making the insertion manager class `static` and see all the red Xs…)

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
```diff
@@ -712,6 +730,90 @@ protected void updateBookkeeping(Record record) throws NoGuidForIdException,
...
+      Uri recordURI = dbHelper.insert(toStore);
+      if (recordURI == null) {
+        delegate.onRecordStoreFailed(new RuntimeException("Got null URI inserting folder with guid " + toStore.guid + "."));
+        return false;
+      }
+      toStore.androidID = ContentUris.parseId(recordURI);
+      Logger.debug(LOG_TAG, "Inserted folder with guid " + toStore.guid + " as androidID " + toStore.androidID);
+
+      try {
+        updateBookkeeping(toStore);
```
rnewman: Ah, OK, so your previous comment is wrong — `AndroidBrowserBookmarksInsertionManager` is not static, so it indirectly modifies the GUID maps via `updateBookkeeping`.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
```diff
@@ -712,6 +730,90 @@ protected void updateBookkeeping(Record record) throws NoGuidForIdException,
...
+    public AndroidBrowserBookmarksInsertionManager(int flushThreshold, Set<String> writtenFolders) {
+      super(flushThreshold, writtenFolders);
+    }
+
+    @Override
+    protected boolean insertFolder(BookmarkRecord record) {
+      // A folder that is *not* deleted needs its androidID updated, so that
+      // updateBookkeeping can re-parent, etc.
+      Record toStore = prepareRecord(record);
+      Uri recordURI = dbHelper.insert(toStore);
```
rnewman: I'd be really inclined to wrap this in a `try` block, too. If this throws an NPE or something on a background thread, sadness will ensue.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
```diff
@@ -712,6 +730,90 @@ protected void updateBookkeeping(Record record) throws NoGuidForIdException,
...
 
+  protected final ArrayList<String> insertedGuids = new ArrayList<String>();
+
+  public class AndroidBrowserBookmarksInsertionManager extends BookmarksInsertionManager {
+    public AndroidBrowserBookmarksInsertionManager(int flushThreshold, Set<String> writtenFolders) {
+      super(flushThreshold, writtenFolders);
+    }
+
+    @Override
+    protected boolean insertFolder(BookmarkRecord record) {
```
rnewman: Why are you not writing into `writtenFolders` here? In fact, what is `writtenFolders` actually for? I don't see anywhere that it gets used.

(EDIT): ah, it's `insertedFolders` elsewhere. So: why are we adding to `insertedGuids`, and calling `updateBookkeeping`, but not adding to the insertion manager's `insertedFolders`?

(EDIT AGAIN): ah, it's done by the superclass's recursive add method. Confuzzling in a linear diff.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
```diff
@@ -549,14 +557,24 @@ public void begin(RepositorySessionBeginDelegate delegate) throws InvalidSession
       cur.close();
     }
     deletionManager = new BookmarksDeletionManager(dataAccessor, DEFAULT_DELETION_FLUSH_THRESHOLD);
+
+    // We just crawled the database enumerating all folders; we'll start the
+    // insertion manager with exactly these folders as the known parents. From
+    // here on, there is no reference to <code>parentGuidToIDMap</code> in the
+    // insertion manager.
+    Set<String> guidsKnownBeforeInsertions = new HashSet<String>(parentGuidToIDMap.keySet());
```
rnewman: You don't need to allocate a new set here: `BookmarksInsertionManager` copies the contents.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
```diff
@@ -712,6 +730,90 @@ protected void updateBookkeeping(Record record) throws NoGuidForIdException,
...
+
+    @Override
+    protected void bulkInsertNonFolders(Collection<BookmarkRecord> records) {
+      // All of these records are *not* deleted and *not* folders, so we don't
+      // need to update androidID at all!
+      // TODO: persist records that fail to insert for later retry.
+      ArrayList<Record> toStores = new ArrayList<Record>(records.size());
+      for (Record record : records) {
+        Record toStore = (BookmarkRecord) prepareRecord(record);
+        toStores.add(toStore);
```
rnewman: Just inline `toStore`, and you can discard the cast.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
```diff
@@ -725,10 +827,42 @@ protected void storeRecordDeletion(final Record record, final Record existingRec
...
+        Logger.warn(LOG_TAG, "Enqueued but didn't insert the following guids: " +
+            "(" + Utils.toCommaSeparatedString(eSet) + ")");
+      }
+
+      iSet.removeAll(new HashSet<String>(enqueuedGuids));
+      if (iSet.isEmpty()) {
+        // Should always be the case.
+      } else {
+        Logger.warn(LOG_TAG, "Inserted but didn't enqueue the following guids: " +
+            "(" + Utils.toCommaSeparatedString(iSet) + ")");
```
rnewman: This pile of code allocates three sets of fifty strings each. That's a pile of memory that might nudge us into OOM on some devices. Remember that the entire process has a 16MB heap limit.

I suggest killing `insertedGuids` and being more aggressive about designing for a constrained-memory device. We can't recover from this situation, so why not use two counters?

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserHistoryRepositorySession.java
```diff
@@ -148,17 +148,11 @@ public void finish(final RepositorySessionFinishDelegate delegate) throws Inacti
...
-   *         <code>androidID</code> is not valid! It will be set after the
-   *         records are flushed to the database.
    */
   @Override
-  protected Record insert(Record record) throws NoGuidForIdException, NullCursorException, ParentNotFoundException {
+  protected void insert(Record record) throws NoGuidForIdException, NullCursorException, ParentNotFoundException {
     HistoryRecord toStore = (HistoryRecord) prepareRecord(record);
-    toStore.androidID = -111; // Hopefully this special value will make it easy to catch future errors.
-    updateBookkeeping(toStore); // Does not use androidID -- just GUID -> String map.
     enqueueNewRecord(toStore);
```
rnewman: Again, inline `toStore`.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserRepositoryDataAccessor.java
```diff
@@ -177,4 +179,40 @@ public void updateByGuid(String guid, ContentValues cv) {
...
+  public int bulkInsert(List<Record> records) throws NullCursorException {
+    if (records.isEmpty()) {
+      Logger.debug(LOG_TAG, "No records to insert, returning.");
+    }
+
+    int size = records.size();
+    ContentValues[] cvs = new ContentValues[size];
+    int index = 0;
+    for (Record record : records) {
+      cvs[index] = getContentValues(record);
```
rnewman: Note that `getContentValues` can throw a runtime exception, for bookmarks at least… think about how you want to handle that.

src/main/java/org/mozilla/gecko/sync/repositories/android/BookmarksInsertionManager.java
```diff
@@ -0,0 +1,292 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+package org.mozilla.gecko.sync.repositories.android;
```
rnewman, line 5: Newline.

src/main/java/org/mozilla/gecko/sync/repositories/android/BookmarksInsertionManager.java
```diff
@@ -0,0 +1,292 @@
...
+    if (num < flushThreshold) {
+      Logger.debug(LOG_TAG, "Incremental flush called with " + num + " < " + flushThreshold + " non-folders; not flushing.");
+      return;
+    }
+    Logger.debug(LOG_TAG, "Incremental flush called with " + num + " non-folders; flushing in batches of " + flushThreshold);
+
+    // bulkInsert non folders in batches.  First copy to array, then slice up array, then re-insert into ordered set.
+    ArrayList<BookmarkRecord> toWrite = new ArrayList<BookmarkRecord>(nonFoldersToWrite);
+    int beg = 0;
+    int end = flushThreshold;
```
rnewman: The purpose of incremental flushing is to get stuff out of memory and on to disk. I don't see any point in bulkInserting only some of the stuff we have queued up, particularly if we are calling `incrementalFlush` on every possible insertion. The only time we'll be above `flushThreshold` is if we queue up a ton of records and then find their parent, and that's (a) pretty unlikely, and (b) we're probably better off flushing the whole lot rather than allocating a new `ArrayList` and a bunch of subList references.

src/main/java/org/mozilla/gecko/sync/repositories/android/BookmarksInsertionManager.java
```diff
@@ -0,0 +1,292 @@
...
+    while (end <= num) {
+      List<BookmarkRecord> batch = toWrite.subList(beg, end);
+      bulkInsertNonFolders(batch);
+      beg += flushThreshold;
+      end += flushThreshold;
+    }
+    nonFoldersToWrite.clear();
+    nonFoldersToWrite.addAll(toWrite.subList(beg, num));
```
rnewman: This is an inefficient way to do this, sadly. You're using progressively _more_ memory on each batch.

`subList` does not do what you think it does. It retains a reference to the original `List`, and a little more memory besides. And you're chaining them.

http://teddziuba.com/2008/09/java-sublist-gotcha.html

Give up on all this: just call `bulkInsertNonFolders(nonFoldersToWrite); nonFoldersToWrite = new …`, or `clear` if you prefer.

src/main/java/org/mozilla/gecko/sync/repositories/android/BookmarksInsertionManager.java
```diff
@@ -0,0 +1,292 @@
...
+    // the non-folders for bulk insertion.
+    int numFolders = 0;
+    int numNonFolders = 0;
+    for (Set<BookmarkRecord> records : recordsWaitingForParent.values()) {
+      for (BookmarkRecord record : records) {
+        if (record.isFolder()) {
+          numFolders += 1;
+          if (!insertFolder(record)) {
+            Logger.warn(LOG_TAG, "Folder with known parent with guid " + record.parentID + " failed to insert!");
+            return;
```
rnewman: You really want to `return` here? I think you mean `continue`.

src/main/java/org/mozilla/gecko/sync/repositories/android/BookmarksInsertionManager.java
```diff
@@ -0,0 +1,292 @@
...
+          }
+          Logger.debug(LOG_TAG, "Folder with known parent with guid " + record.parentID + " inserted; adding to inserted folders.");
+          insertedFolders.add(record.guid);
+          continue;
+        }
```
rnewman, line 224: This is one of the few times I'll suggest using an `else`. A mess of `continue` and `break` and `return` is worse than a little structured programming.

src/main/java/org/mozilla/gecko/sync/repositories/android/BookmarksInsertionManager.java
```diff
@@ -0,0 +1,292 @@
...
+    }
+    recordsWaitingForParent.clear();
+    incrementalFlush(); // Write as many non-folder batches as possible, and then whatever is left.
+    if (!nonFoldersToWrite.isEmpty()) {
+      bulkInsertNonFolders(nonFoldersToWrite);
+      nonFoldersToWrite.clear();
+    }
+
+    Logger.debug(LOG_TAG, "finishUp inserted " + numFolders + " folders without known parents " +
+        "and " + numNonFolders + " non-folders without known parents.");
```
rnewman: Fix this indenting, please.

test/src/org/mozilla/android/sync/test/AndroidBrowserBookmarksRepositoryTest.java
```diff
@@ -605,4 +607,33 @@ public void testSqlInjectDelete() {
...
+    ArrayList<Record> recordList = new ArrayList<Record>();
+    recordList.add(bookmark1);
+    recordList.add(bookmark2);
+    db.bulkInsert(recordList);
+
+    String[] guids = new String[] { bookmark1.guid, bookmark2.guid };
+    Record[] expected = new Record[] { bookmark1, bookmark2 };
+    performWait(fetchRunnable(session, guids, expected));
+    dispose(session);
+  }
```
rnewman, line 638: Please also add a test here where we expect to replace a local folder with a new folder, whilst adding children to it.

I want to make sure that `replace` still functions under batching.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
rnewman: I believe we can count on a `wipe` being followed by a `finish`. So this is just for cleanup.

rnewman:
Looking pretty good. Remaining comments essentially fall into "worried about replace; worried about too much allocation; little safety and correctness gotchas". So a test and deleting a little bit of code and we should be good to go.

rnewman:
Rebased and merged.

PR #173, opened 2012-04-24
Bug 747065 - Precondition failed on clients PUT

rnewman:
Rewrite of Marina's branch. Fixed tests. Apparently the header/body mismatch does not occur for POST, so this should be fine.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/GlobalConstants.java.in
+++ b/src/main/java/org/mozilla/gecko/sync/GlobalConstants.java.in
@@ -10,7 +10,7 @@ package org.mozilla.gecko.sync;
  */
 public class GlobalConstants {
   public static final String PRODUCT_NAME = "@MOZ_APP_DISPLAYNAME@";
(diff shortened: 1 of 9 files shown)


rnewman:
This has not been thoroughly tested; automated tests only.

ncalexan:
- these commits cannot be "rolled back" individually and still compile (let alone pass tests!).  This worries me but is not a show stopper.
- please move final "bump version string" commit into its own branch with its own merge commit, and say "bump version string from 0.6 to 0.7".  Let's make it easy to figure out when this happened.
- I notice you dropped exposing the request in SyncStorageResponse.  You argued that it was necessary and useful once...

r+ after nits.

rnewman:
- Atomic commits: that's fine, we don't bisect into merge commits. They're separate for review. They'll all be a single commit in hg.
- Version change won't be a merge commit, but it will be its own bug. It's just there so I don't forget to land it with this change.
- Accessing the request that way was the right way to introspect the method. We no longer need to do that, as you can see — we use POST instead — so it's gone entirely.

PR #176, opened 2012-04-28
Bug 749857 - Don't upload history records without URI or visits.

rnewman:


rnewman:
Test will follow.

rnewman:
Merged with test.

PR #177, opened 2012-05-01
Bug 714304 - Implement handling of "upgrade required" response.

liuche:
Adds handling of HTTP 400: Upgrade Required responses from server in EnsureClusterURLStage.

On receiving an "upgrade required," disable automatic syncing and sync accounts and disable syncing of accounts. On reinstall of the Fennec + Sync package, the sync accounts are re-enabled and set to automatically sync.

This may require changes once we move to supporting multiple Firefox versions, because the sync accounts will still be set to not sync automatically, and be disabled.

Feedback requested, not ready for merging with develop because of mvn test dependencies for MockServer et al.

### DIFF
--- a/manifests/SyncAndroidManifest_activities.xml.in
+++ b/manifests/SyncAndroidManifest_activities.xml.in
@@ -18,3 +18,9 @@
             android:name="org.mozilla.gecko.sync.setup.activities.SetupFailureActivity" />
         <activity
             android:name="org.mozilla.gecko.sync.setup.activities.SetupSuccessActivity" />
+        <receiver
+            android:name="org.mozilla.gecko.sync.receivers.UpgradeReceiver">
+            <intent-filter>
+                <action android:name="android.intent.action.MY_PACKAGE_REPLACED" />
+            </intent-filter>
+        </receiver>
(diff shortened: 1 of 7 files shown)


rnewman:
Need two changes in logic here to avoid some pain.

Firstly, we should hook in this error code detection into the place where we _already_ have error code detection. We even already have a callback path back to a place that's aware of our Account object! See comments.

Secondly, I'm not super keen on unconditionally turning syncing off and on for an account.

If I have two sync accounts, and I receive an Upgrade Required response on one, I certainly don't want to disable them both. (And does Android allow me to force a sync on an account that's unchecked?)

Or: what if I set up Sync, disable the account, then upgrade Fennec? Is my account still disabled? I sure hope so, because I'm roaming in Egypt next week and it costs me $10/KB.

So: we should check to see if our Account is enabled, and that we got an Upgrade Required response, and set a pref somewhere or a flag on the account that indicates that we're turning it off because an upgrade is required. When we get upgraded, only re-enable accounts that have this flag set.

Finally, when we tell `SyncAdapter` that we've had a problem, it should _probably_ increment `syncResult.stats.numConflictDetectedExceptions`. The only reason _not_ to do that is if we can't undo the hard error by setting `setIsSyncable`.

Which of `setIsSyncable` and `setSyncAutomatically` controls the checkbox in Settings? If that's whether to sync automatically, we're safe in calling `setIsSyncable` every which way, and allowing the user preference to override. Otherwise, we need to be much more nuanced.

Apologies for the rambling ill-defined review!

liuche:
Updated to #181.

PR #180, opened 2012-05-04
Bug 750346 - Persist engine enabled state, process meta/global engine contents.

rnewman:
Extracted from Nick's megabug.


rnewman:
Done.

PR #181, opened 2012-05-04
Bug 714304 - Implement handling of "upgrade required" response.

liuche:
Addressed comments made in pull request #177.

Also decoupled testing from MockServer and HTTPTestServerHelper, but request hangs and doesn't complete. Feedback requested for TestUpgradeRequired.

Requiring an upgrade will uncheck the checkboxes in the account settings if checked, and set a flag for the accounts that have been changed this way. (AFAICT, there isn't a way to filter by accounts associated with a particular Authenticator.)

Upgrading the package will check the checkboxes for accounts that have a flag and clear the flag.

Note: setSyncAutomatically controls checkbox behavior, setIsSyncable controls whether a Firefox installation associated with an account appears at all in Account Settings.


liuche:
Realized mvn test was failing to compile or find some Mock\* classes, added fixes.

rnewman:
Adjusted and merged.

PR #184, opened 2012-05-08
Bug 738676 - Pair a device not working a second time.

liuche:
Problem had to do with multiple instances of Settings running, and only one instance being able to successfully start the Pair activity through intents. Maybe Android doesn't expect multiple Settings instances to run, and starting intents from the Preferences xml of our Sync account didn't work.

Repro-d bug on my Asus Transformer, no longer appears after fix. Hopefully this is not a device-specific fix (again). QA on different devices will definitely be necessary.


liuche:
Specifically, would like tchung to test, because some previous irc convos seemed to indicate that the Settings task/activity stack was not the problem on his device.

PR #185, opened 2012-05-09
generate manifests used for Mozilla build system with one entry per line

jhford:
Part of mozilla bug 752873, instead of having the tr overhead on every android build, move the work of generating one file per manifest file line to the android-sync code drop script.  Another approach is to generate a makefile that has the date from the manifests.  This avoids the shell call overhead entirely.

### DIFF
--- a/fennec-copy-code.sh
+++ b/fennec-copy-code.sh
@@ -48,8 +48,9 @@ done
 echo "Writing README."
 echo $WARNING > $ANDROID/base/sync/README.txt
 echo $WARNING > $ANDROID/base/httpclientandroidlib/README.txt
+true > $SYNC/java-sources.mn
 for f in $SOURCEFILES ; do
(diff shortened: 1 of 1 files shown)


jhford:
The first two commits in the pull request would require landing https://bug752873.bugzilla.mozilla.org/attachment.cgi?id=622419 and the makefile would require https://bug752873.bugzilla.mozilla.org/attachment.cgi?id=622424

https://tbpl.mozilla.org/?tree=Try&rev=d70761ec02ed .  The logs for a green build with the test push are here http://ftp.mozilla.org/pub/mozilla.org/firefox/try-builds/jford@mozilla.com-d70761ec02ed/try-android/try-android-bm14-try1-build1657.txt.gz

rnewman:
Manually merged with fixes.

PR #190, opened 2012-05-11
Bug 752514 - fix comment in SyncAccounts.

ncalexan:
Not a code change, but caused 752514.


ncalexan:
I considered validating sync keys, but I'm not convinced we can assume a 26 char alpha-num key, especially not for profile migrating accounts.

PR #192, opened 2012-05-16
Bug 755576 - Bookmark cycle errors.

rnewman:


src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserBookmarksRepositorySession.java
```diff
@@ -298,19 +298,20 @@ private String getParentName(String parentGUID) throws ParentNotFoundException,
    * @param persist
    *        True if generated positions should be written to the database. The modified
    *        time of the parent folder is only bumped if this is true.
+   * @param childArray
+   *        A JSONArray which will be populated with an array of GUIDs.
    * @return
-   *        An array of GUIDs.
+   *        True if the resulting array is "clean" (i.e., reflects the content of the database).
```
ncalexan: nit: you don't actually clear `childArray`, so if the caller gives you a non-empty array this will lie.

src/main/java/org/mozilla/gecko/sync/repositories/Server11RepositorySession.java
```diff
@@ -359,9 +359,21 @@ protected void flush() {
 
   @Override
   public void storeDone() {
+    Logger.debug(LOG_TAG, "storeDone().");
     synchronized (recordsBufferMonitor) {
       flush();
-      storeDone(uploadTimestamp.get());
+      // Do this in a Runnable so that the timestamp is grabbed after any upload.
```
ncalexan, line 365: This is clearly correct and should be merged ASAP, but it's hard to say that it will fix Bug 755576.

rnewman, line 365: According to my testing, it does: uploaded records now correctly fall into the window between last fetch and last store, and are thus skipped on the next sync.

ncalexan:
r=nalexander after fixing above nit about clearing array.

rnewman:
Done and merged: 9d28c21.

PR #193, opened 2012-05-16
Bug 754958 - Pairing aborts if device screen turns off.

liuche:
Added flags to keep device screen on. Added the flag to AccountActivity as well, because launching and canceling AccountActivity quickly sometimes led to the "screen on" flag not getting applied to the J-PAKE setup activity.


ncalexan:
r=nalexander

PR #194, opened 2012-05-16
Return focus to fennec after cancelled Sync Setup.

liuche:
Finish J-PAKE setup screen if manual setup completed successfully. While testing a different fix, I found that one can return to the J-PAKE screen after successfully doing manual setup.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/setup/activities/SetupSyncActivity.java
+++ b/src/main/java/org/mozilla/gecko/sync/setup/activities/SetupSyncActivity.java
@@ -598,7 +598,7 @@ public class SetupSyncActivity extends AccountAuthenticatorActivity {
 
   @Override
   public void onActivityResult(int requestCode, int resultCode, Intent data) {
-    switch(resultCode) {
+    switch (resultCode) {
     case Activity.RESULT_OK:
       // Setup completed in manual setup.
       finish();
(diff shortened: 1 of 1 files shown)


src/main/java/org/mozilla/gecko/sync/setup/activities/SetupSyncActivity.java
```diff
@@ -586,4 +587,13 @@ public void run() {
       }
     });
   }
+
+  @Override
+  public void onActivityResult(int requestCode, int resultCode, Intent data) {
+    switch(resultCode) {
```
rnewman: Space before parens.

src/main/java/org/mozilla/gecko/sync/setup/activities/SetupSyncActivity.java
```diff
@@ -174,7 +175,7 @@ public void onDestroy() {
   public void manualClickHandler(View target) {
     Intent accountIntent = new Intent(this, AccountActivity.class);
     accountIntent.setFlags(Intent.FLAG_ACTIVITY_NO_ANIMATION);
-    startActivity(accountIntent);
+    startActivityForResult(accountIntent, 0);
```
rnewman, line 186: Should that `0` be some named value like `Activity.RESULT_OK`? Magic numbers…

rnewman:
Bug number and appropriate commit message, please.

src/main/java/org/mozilla/gecko/sync/setup/activities/SetupSyncActivity.java
liuche, line 186: That 0 is a requestCode, and is passed to the handler to differentiate between multiple activities that were started with startActivityForResult. It just has to be >= 0, and since we only have one case, I just used 0. It can be made a constant, something like ACTIVITY_SETUP_MANUAL.

rnewman:
Merged as 076405b.

PR #195, opened 2012-05-17
Bug 755935 - Regression: J-PAKE code is refreshed on device rotation.

liuche:
New configChanges parameter, screenSize, added in ICS.

http://developer.android.com/reference/android/R.attr.html#configChanges

Wasn't handled before.


ncalexan:
r=nalexander

PR #197, opened 2012-05-21
Bug 756477 - Flush log level cache on each sync to detect log level changes

rnewman:


ncalexan:
r=nalexander.  I see a mix of `Log` and `Logger` in `SyncAdapter`; some is intentional, but some should probably be changed over.  Perhaps we should add `Logger.force(LOG_TAG, level, message)` or similar?

rnewman:
I like that idea; file a bug!

PR #199, opened 2012-05-23
Bug 736393 - no abort on store failure.

ncalexan:
This... got ugly.  I tried to take a "minimal change" path through the existing code, but I wouldn't be surprised if rnewman had an entirely different idea of how this was supposed to work in the failure cases.

Enjoy some actual tests, though!

Device testing to come tomorrow AM, when m-i finally finishes building.


ncalexan:
Device testing looks about as good as it can.  I disconnect wifi during downloads and see good things -- failure on fetch, correctly caught and error indicated.  I disconnect wifi during upload and see upload errors correctly caught and sync aborted.

rnewman:
This looks good to me. If this doesn't fall down for your own device testing, I'd like to see this bake on m-i/m-c, and request uplift on Monday.

r=me

PR #206, opened 2012-06-04
Bug 760614 - Retry on connection failures, to address transient network errors.

liuche:
Retry 5 times on an I/O error. We may want to increase this limit, or pause between retries, depending on how quickly we think networks will recover.

Still see exception logging from httpclient on the I/O failure with the SSL pipe, further upstream.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/net/BaseResource.java
+++ b/src/main/java/org/mozilla/gecko/sync/net/BaseResource.java
@@ -61,7 +61,7 @@ public class BaseResource implements Resource {
   private static final int MAX_TOTAL_CONNECTIONS     = 20;
   private static final int MAX_CONNECTIONS_PER_ROUTE = 10;
 
-  private boolean retry = true;
+  private boolean retryOnFailedRequest = true;
 
   public static boolean rewriteLocalhost = true;
 
@@ -73,7 +73,6 @@ public class BaseResource implements Resource {
   public    ResourceDelegate delegate;
   protected HttpRequestBase request;
   public String charset = "utf-8";
-
 
   protected static WeakReference<HttpResponseObserver> httpResponseObserver = null;
 
@@ -253,24 +252,27 @@ public class BaseResource implements Resource {
       delegate.handleHttpProtocolException(e);
     } catch (IOException e) {
       Logger.debug(LOG_TAG, "I/O exception returned from execute.");
-      if (!retry) {
+      if (!retryOnFailedRequest) {
         delegate.handleHttpIOException(e);
       } else {
-        retry = false;
-        Logger.debug(LOG_TAG, "Retrying request...");
-        this.execute();
+        retryRequest();
       }
     } catch (Exception e) {
       // Bug 740731: Don't let an exception fall through. Wrapping isn't
       // optimal, but often the exception is treated as an Exception anyway.
-      if (!retry) {
+      if (!retryOnFailedRequest) {
         delegate.handleHttpIOException(new IOException(e));
       } else {
-        retry = false;
-        Logger.debug(LOG_TAG, "Retrying request...");
-        this.execute();
+        retryRequest();
       }
     }
+  }
+
+  private void retryRequest() {
+    // Only retry once.
+    retryOnFailedRequest = false;
+    Logger.debug(LOG_TAG, "Retrying request...");
+    this.execute();
   }
 
   private void go(HttpRequestBase request) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/mozilla/gecko/sync/net/BaseResource.java
```diff
@@ -249,8 +252,32 @@ private void execute() {
...
+        this.post(((HttpPost) request).getEntity());
+      } else if ("PUT".equals(httpMethod)) {
+        this.post(((HttpPut) request).getEntity());
+      } else {
+        // Unknown request type.
+        Logger.debug(LOG_TAG, "Retry not handled; unknown request type " + httpMethod);
+        delegate.handleHttpIOException(e);
+      }
     } catch (Exception e) {
+      Logger.debug(LOG_TAG, "Catching general exception.");
```
ncalexan: I don't think this will work.  The error we were not catching (from Bug 740731) is

java.lang.IllegalStateException: Connection is not open at ch.boye.httpclientandroidlib.impl.SocketHttpClientConnection.assertOpen(SocketHttpClientConnection.java)

I thought that was the indicator that we needed to try again, right?

src/main/java/org/mozilla/gecko/sync/net/BaseResource.java
```diff
@@ -72,6 +72,9 @@
   protected HttpRequestBase request;
   public String charset = "utf-8";
 
+  private static final int MAX_RETRIES = 5;
```
rnewman: I am not particularly keen on this many retries.

If we hit the same exception on the second retry, there's no point hammering on our error logs. I'd be inclined to try exactly once more, unless we see real evidence that suggests that we need more.

src/main/java/org/mozilla/gecko/sync/net/BaseResource.java
```diff
@@ -249,14 +252,53 @@ private void execute() {
...
+      this.post(((HttpPost) request).getEntity());
+    } else if ("PUT".equals(httpMethod)) {
+      this.post(((HttpPut) request).getEntity());
+    } else {
+      // Unknown request type.
+      Logger.debug(LOG_TAG, "Retry not handled; unknown request type " + httpMethod);
+      return false;
+    }
+    return true;
+  }
```
ncalexan: nit: newline.

src/main/java/org/mozilla/gecko/sync/net/BaseResource.java
```diff
@@ -249,14 +252,53 @@ private void execute() {
...
+    if (numRetries++ >= MAX_RETRIES) {
+      Logger.debug(LOG_TAG, "Exceeded retry limit.");
+      return false;
+    }
+
+    // Retry request.
+    Logger.debug(LOG_TAG, "Retrying request...");
+    String httpMethod = request.getMethod();
+    if ("GET".equals(httpMethod)) {
+      this.get();
```
rnewman: Why aren't you just re-executing? Does that not work (because of entity reuse?)?

liuche: Er, no reason, my mistake. Fixed that in the newest commit - hand-tested reusing the entity, with no problems.

src/main/java/org/mozilla/gecko/sync/net/BaseResource.java
```diff
@@ -72,6 +74,7 @@
   protected HttpRequestBase request;
   public String charset = "utf-8";
 
+
```
ncalexan: nit: blank line.

src/main/java/org/mozilla/gecko/sync/net/BaseResource.java
```diff
@@ -61,6 +61,8 @@
   private static final int MAX_TOTAL_CONNECTIONS     = 20;
   private static final int MAX_CONNECTIONS_PER_ROUTE = 10;
 
+  private boolean retry = true;
```
ncalexan: Either choose a descriptive name, or include a javadoc dis-ambiguating.

src/main/java/org/mozilla/gecko/sync/net/BaseResource.java
```diff
@@ -249,11 +252,24 @@ private void execute() {
     } catch (ClientProtocolException e) {
       delegate.handleHttpProtocolException(e);
     } catch (IOException e) {
-      delegate.handleHttpIOException(e);
+      Logger.debug(LOG_TAG, "I/O exception returned from execute.");
```
ncalexan, line 254: This looks good.

ncalexan:
r=nalexander after addressing comments.

ncalexan:
liuche, this would close automatically if you pushed `liuche/retry`: I think github is clever enough to realize when a branch has been merged and close the pull req.

PR #207, opened 2012-06-04
Bug 714657 - Persist earliest next sync in Android Account data rather t...

ncalexan:
...han "sync.prefs.global" Shared Preferences.

Long term sync.prefs.globals should go away, and certainly next sync
is Account specific, so this seemed the right thing to do.  I made
next sync an Account datum because we want to access it without
invoking the full Account Manager callback needed to retrieve encoded
username and server URL.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/Utils.java
+++ b/src/main/java/org/mozilla/gecko/sync/Utils.java
@@ -176,22 +176,61 @@ public class Utils {
     return (long)(decimal * 1000);
   }
 
-  public static byte[] sha1(String utf8)
+  protected static byte[] sha1(String utf8)
       throws NoSuchAlgorithmException, UnsupportedEncodingException {
     MessageDigest sha1 = MessageDigest.getInstance("SHA-1");
     return sha1.digest(utf8.getBytes("UTF-8"));
   }
 
-  public static String sha1Base32(String utf8)
+  protected static String sha1Base32(String utf8)
       throws NoSuchAlgorithmException, UnsupportedEncodingException {
     return new Base32().encodeAsString(sha1(utf8)).toLowerCase(Locale.US);
   }
 
-  public static String getPrefsPath(String username, String serverURL)
-    throws NoSuchAlgorithmException, UnsupportedEncodingException {
-    return "sync.prefs." + sha1Base32(serverURL + ":" + username);
+  /**
+   * If we encounter characters not allowed by the API (as found for
+   * instance in an email address), hash the value.
+   * @param account
+   *        An account string.
+   * @return
+   *        An acceptable string.
+   * @throws UnsupportedEncodingException
+   * @throws NoSuchAlgorithmException
+   */
+  public static String usernameFromAccount(String account) throws NoSuchAlgorithmException, UnsupportedEncodingException {
+    if (account == null || account.equals("")) {
+      throw new IllegalArgumentException("No account name provided.");
+    }
+    if (account.matches("^[A-Za-z0-9._-]+$")) {
+      return account.toLowerCase(Locale.US);
+    }
+    return sha1Base32(account.toLowerCase(Locale.US));
   }
 
+  /**
+   * Get shared preferences path for a Sync account.
+   *
+   * @param username the Sync account name, optionally encoded with <code>Utils.usernameFromAccount</code>.
+   * @param serverURL the Sync account server URL.
+   * @return the path.
+   * @throws NoSuchAlgorithmException
+   * @throws UnsupportedEncodingException
+   */
+  public static String getPrefsPath(String username, String serverURL)
+    throws NoSuchAlgorithmException, UnsupportedEncodingException {
+    return "sync.prefs." + sha1Base32(serverURL + ":" + usernameFromAccount(username));
+  }
+
+  /**
+   * Get shared preferences for a Sync account.
+   *
+   * @param context Android <code>Context</code>.
+   * @param username the Sync account name, optionally encoded with <code>Utils.usernameFromAccount</code>.
+   * @param serverURL the Sync account server URL.
+   * @return a <code>SharedPreferences</code>.
+   * @throws NoSuchAlgorithmException
+   * @throws UnsupportedEncodingException
+   */
   public static SharedPreferences getSharedPreferences(Context context, String username, String serverURL) throws NoSuchAlgorithmException, UnsupportedEncodingException {
     String prefsPath = getPrefsPath(username, serverURL);
     Logger.debug(LOG_TAG, "Shared preferences: " + prefsPath);
--- a/src/main/java/org/mozilla/gecko/sync/Utils.java
+++ b/src/main/java/org/mozilla/gecko/sync/Utils.java
@@ -176,13 +176,13 @@ public class Utils {
     return (long)(decimal * 1000);
   }
 
-  protected static byte[] sha1(String utf8)
+  protected static byte[] sha1(final String utf8)
       throws NoSuchAlgorithmException, UnsupportedEncodingException {
     MessageDigest sha1 = MessageDigest.getInstance("SHA-1");
     return sha1.digest(utf8.getBytes("UTF-8"));
   }
 
-  protected static String sha1Base32(String utf8)
+  protected static String sha1Base32(final String utf8)
       throws NoSuchAlgorithmException, UnsupportedEncodingException {
     return new Base32().encodeAsString(sha1(utf8)).toLowerCase(Locale.US);
   }
@@ -197,7 +197,7 @@ public class Utils {
    * @throws UnsupportedEncodingException
    * @throws NoSuchAlgorithmException
    */
-  public static String usernameFromAccount(String account) throws NoSuchAlgorithmException, UnsupportedEncodingException {
+  public static String usernameFromAccount(final String account) throws NoSuchAlgorithmException, UnsupportedEncodingException {
     if (account == null || account.equals("")) {
       throw new IllegalArgumentException("No account name provided.");
     }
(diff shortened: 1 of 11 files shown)


rnewman:
Not quite true.

Account should hold attributes that are specific to the account: the sync key, the server.

Prefs holds attributes that are specific to some (account, fennec, …) tuple. This is one such thing.

Consider if the same Android account is syncing with Nightly and Beta. By storing the next sync timestamp in the account, you're ensuring that only one of them will get to sync.

In general we should be using prefs for anything that's not "pretty much immutable".

ncalexan:
I'm not sure I agree with you re: what should be stored in Account and what should not, but I really don't have full picture re: multiple Fennecs, multiple profiles, etc so I won't move on that now.

I've rewritten this to correctly wipe the user prefs (there were encoding bugs) and to wipe the global prefs too (since we only support one account right now, we can always wipe the global prefs on new account creation).  Device tests are good for me.

I would like to verify this doesn't break migrating settings during profile migration (cluster URL and sync ID), and I am confident it won't, but my Fennec is borked so it will have to wait.

src/main/java/org/mozilla/gecko/sync/Utils.java
```diff
@@ -176,22 +176,61 @@ public static long decimalSecondsToMilliseconds(Integer decimal) {
     return (long)(decimal * 1000);
   }
 
-  public static byte[] sha1(String utf8)
+  protected static byte[] sha1(String utf8)
```
rnewman: While we're touching this, can we make `utf8` `final`?

src/main/java/org/mozilla/gecko/sync/Utils.java
```diff
@@ -176,22 +176,61 @@ public static long decimalSecondsToMilliseconds(Integer decimal) {
...
       throws NoSuchAlgorithmException, UnsupportedEncodingException {
     MessageDigest sha1 = MessageDigest.getInstance("SHA-1");
     return sha1.digest(utf8.getBytes("UTF-8"));
```
rnewman: Curse you, Java, and your ridiculous checked exceptions for built-in encodings.

src/main/java/org/mozilla/gecko/sync/Utils.java
```diff
@@ -176,22 +176,61 @@ public static long decimalSecondsToMilliseconds(Integer decimal) {
...
   }
 
-  public static String sha1Base32(String utf8)
+  protected static String sha1Base32(String utf8)
```
rnewman: Also `final`.

src/main/java/org/mozilla/gecko/sync/Utils.java
```diff
@@ -176,22 +176,61 @@ public static long decimalSecondsToMilliseconds(Integer decimal) {
...
+   * If we encounter characters not allowed by the API (as found for
+   * instance in an email address), hash the value.
+   * @param account
+   *        An account string.
+   * @return
+   *        An acceptable string.
+   * @throws UnsupportedEncodingException
+   * @throws NoSuchAlgorithmException
+   */
+  public static String usernameFromAccount(String account) throws NoSuchAlgorithmException, UnsupportedEncodingException {
```
rnewman: `final`.

src/main/java/org/mozilla/gecko/sync/Utils.java
```diff
@@ -176,22 +176,61 @@ public static long decimalSecondsToMilliseconds(Integer decimal) {
...
+    if (account.matches("^[A-Za-z0-9._-]+$")) {
+      return account.toLowerCase(Locale.US);
+    }
+    return sha1Base32(account.toLowerCase(Locale.US));
+  }
+
+  /**
+   * Get shared preferences path for a Sync account.
+   *
+   * @param username the Sync account name, optionally encoded with <code>Utils.usernameFromAccount</code>.
```
rnewman, line 213: You need to alter this comment if you're changing the behavior. Previously it was the caller's responsibility to hash the username if desired.

src/main/java/org/mozilla/gecko/sync/Utils.java
```diff
@@ -176,22 +176,61 @@ public static long decimalSecondsToMilliseconds(Integer decimal) {
...
+    return "sync.prefs." + sha1Base32(serverURL + ":" + usernameFromAccount(username));
   }
 
+  /**
+   * Get shared preferences for a Sync account.
+   *
+   * @param context Android <code>Context</code>.
+   * @param username the Sync account name, optionally encoded with <code>Utils.usernameFromAccount</code>.
+   * @param serverURL the Sync account server URL.
+   * @return a <code>SharedPreferences</code>.
```
rnewman: … instance.

src/main/java/org/mozilla/gecko/sync/Utils.java
```diff
@@ -176,22 +176,61 @@ public static long decimalSecondsToMilliseconds(Integer decimal) {
...
     throws NoSuchAlgorithmException, UnsupportedEncodingException {
-    return "sync.prefs." + sha1Base32(serverURL + ":" + username);
+    return "sync.prefs." + sha1Base32(serverURL + ":" + usernameFromAccount(username));
   }
 
+  /**
+   * Get shared preferences for a Sync account.
+   *
+   * @param context Android <code>Context</code>.
+   * @param username the Sync account name, optionally encoded with <code>Utils.usernameFromAccount</code>.
```
rnewman, line 228: Again, comment is inaccurate.

src/main/java/org/mozilla/gecko/sync/setup/SyncAccounts.java
```diff
@@ -214,9 +214,11 @@ public static Account createSyncAccount(SyncAccountParameters syncAccount) {
     // TODO: for each, also add to res/xml to make visible in account settings
     Logger.debug(LOG_TAG, "Finished setting syncables.");
 
-    // TODO: correctly implement Sync Options.
-    Logger.info(LOG_TAG, "Clearing preferences for this account.");
+    Logger.info(LOG_TAG, "Clearing global prefs.");
+    SyncAdapter.purgeGlobalPrefs(context);
```
rnewman: Hold on sunshine. I'm not sure you ever want to do this.

What global preferences do you think should be cleared when an account is created?

If there are any, should they not instead be local/per-account prefs?

If not, don't we need to have two classes of global prefs?

src/main/java/org/mozilla/gecko/sync/syncadapter/SyncAdapter.java
```diff
@@ -68,25 +68,31 @@ public SyncAdapter(Context context, boolean autoInitialize) {
...
-  private SharedPreferences getGlobalPrefs() {
-    return mContext.getSharedPreferences("sync.prefs.global", SHARED_PREFERENCES_MODE);
+  public static SharedPreferences getGlobalPrefs(Context context) {
+    return context.getSharedPreferences("sync.prefs.global", SHARED_PREFERENCES_MODE);
+  }
+
+  public static void purgeGlobalPrefs(Context context) {
+    getGlobalPrefs(context).edit().clear().commit();
+    final SharedPreferences prefs = context.getSharedPreferences("sync.prefs", SHARED_PREFERENCES_MODE);
+    prefs.edit().remove("global").commit();
```
rnewman: Why do you need these two lines?

"sync.prefs.global" indicates a _file_. This appears to be fetching the prefs file that's named "sync.prefs" and removing from it the entry "global", which is entirely different to the entries in the "sync.prefs.global" file.

ncalexan: You're right.  I added the `clear().commit()` and forgot to remove the old.

src/main/java/org/mozilla/gecko/sync/setup/SyncAccounts.java
ncalexan: This is strictly a bandaid.  We have a ticket for making prefs make sense, and I don't want to move on it yet.  Right now there are three global prefs, none of which are _actually_ global:

 private static final String  PREFS_EARLIEST_NEXT_SYNC = "earliestnextsync";
 private static final String  PREFS_INVALIDATE_AUTH_TOKEN = "invalidateauthtoken";
 private static final String  PREFS_CLUSTER_URL_IS_STALE = "clusterurlisstale";

For this to work, we need to kill "earliestnextsync".  The others can be killed whenever; they're transient and definitely not valid between account creations.  I can do more work and make these prefs account specific, then wipe them with the rest, but I want to get this ticket done _now_ and worry about sorting out what is global and what is account and what is server specific later, when I understand what multiple Fennecs and profiles actually means.

src/main/java/org/mozilla/gecko/sync/Utils.java
ncalexan, line 213: Previously, there was no comment.  I thought "optional encoded" was pretty clear: since `Utils.usernameFromAccount` is idempotent, we can always pass the input through the encoding.  The source of this bug is errors encoding at the callsite; why not fix any that come later?  (There are not enough calls to worry about performance.)

src/main/java/org/mozilla/gecko/sync/Utils.java
ncalexan, line 228: See previous response.

src/main/java/org/mozilla/gecko/sync/setup/SyncAccounts.java
liuche: Add a comment to make it clear that this is a band-aid, and only acceptable with a single Firefox instance (for code maintainability).

liuche:
Sorry this took a while, this looks fine for band-aiding using global prefs. r+ from me.

Is there a bug for making prefs to support multiple firefox instances?

rnewman:
Please correct the title of this pull req.

rnewman:
Chenxia: Bug 761682 covers reorganizing prefs.

PR #203, opened 2012-05-29
Bug 759548 - file logging

ncalexan:
This surfaces UI for setting log levels (check the Account Settings page), and for Sharing and Deleting log files.  Logs are automatically cleaned.

rnewman: please review file code to make sure we're not hitting the disk unnecessarily.

liuche: please make sure UI bits follow whatever our guidelines are.  I'll build and ask madhava and ibarlow for feedback.

Thanks, chap and chappette!

PR #208, opened 2012-06-06
Bug 742109 - use single log tag

ncalexan:
I extracted and cleaned things from the file logging pull request (which just got a lot smaller).  The patches marked pre include a few odds and ends that may as well be here.  The patch marked post touches tons of files (although mostly testing files) and will rot _very_ quickly.  If we're going to get rid of android.util.Log, let's do it now; if we're not, that patch is at the top of the queue for a reason.  Device testing works well for me.


liuche:
This looks good to me, r+, will follow up with reviewing #203.

PR #209, opened 2012-06-06
Bug 762118 - Ignore readinglist items and folder for now.

rnewman:
https://bugzilla.mozilla.org/show_bug.cgi?id=762118


ncalexan:
Is this waiting on tests?

Also, does `readinglist` need to be rooted in the tree (like the `places` and `mobile` folders)?

rnewman:
You can review this part until I deliver tests.

Rooting in the tree will be part of the actual implementation of reading list sync; this just ensures that we never see them.

ncalexan:
This part is fine, then.  The rooting came up when I started grepping for places and other similar (rooted) things.

rnewman:
Here, have some tests!

ncalexan:
r=nalexander.  Didn't get to landing this today, will try again tomorrow if I'm not beaten to it.

PR #210, opened 2012-06-07
Bug 762627 - no singleton command processor

ncalexan:
I checked a few other memory things, and this cycle was not as bad as I thought.  Still worth breaking, however.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/GlobalSession.java
+++ b/src/main/java/org/mozilla/gecko/sync/GlobalSession.java
@@ -166,45 +166,50 @@ public class GlobalSession implements CredentialsSource, PrefsSource, HttpRespon
     // TODO: data-driven plan for the sync, referring to prepareStages.
   }
 
-  protected void registerCommands() {
+  /**
+   * Register commands this global session knows how to process.
+   * <p>
+   * Re-registering a command overwrites any existing registration.
+   */
+  protected static void registerCommands() {
     final CommandProcessor processor = CommandProcessor.getProcessor();
 
     processor.registerCommand("resetEngine", new CommandRunner(1) {
       @Override
-      public void executeCommand(List<String> args) {
+      public void executeCommand(final GlobalSession session, List<String> args) {
         HashSet<String> names = new HashSet<String>();
         names.add(args.get(0));
-        resetStagesByName(names);
+        session.resetStagesByName(names);
       }
     });
 
     processor.registerCommand("resetAll", new CommandRunner(0) {
       @Override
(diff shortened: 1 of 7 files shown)


liuche:
r+ from me.

ncalexan:
Rebased and re-envisioned, so need re-review.  A single long-lived global `CommandProcessor` in some form is needed to implement Send Tab to Device, so I changed focus from removing the singleton to stopping the singleton from holding on to the global session reference.  To that end, I implemented unregistering the commands that reference the current session from the CommandProcessor, which (this being Java) is a little heavier than we'd really like, but... whatever.

src/main/java/org/mozilla/gecko/sync/GlobalSession.java
```diff
@@ -169,38 +176,39 @@ public GlobalSession(String userAPI,
...
       public void executeCommand(List<String> args) {
         HashSet<String> names = new HashSet<String>();
         names.add(args.get(0));
         wipeStagesByName(names);
       }
-    });
+    }));
 
-    processor.registerCommand("wipeAll", new CommandRunner(0) {
+    registeredCommands.add(processor.registerCommand("wipeAll", new CommandRunner(0) {
```
rnewman: This seems a little janky.

Why not have `CommandRunner` take a `GlobalSession` as an argument when invoked? That way they need never store a reference to it.

Alternatively, make `CommandProcessor` non-static, and discard it altogether at the end of a `GlobalSession`'s lifecycle. That'll drop half of the cycle, and everything'll get cleaned up.

ncalexan: Re: make `CommandProcessor` non-static: we need a command processor at two different times: during clients stage of an actual sync, and during an activity completely unrelated to a sync (such as `SendTabActivity`, but also during other activities yet to be developed, such as `WipeOtherClientsActivity`.)  We could have each activity create a command processor, but this seems unnecessary and raises locking and synchronization questions (since all processors access the same clients DB, for example).  A global long-lived singleton seems right to me.

Re: have `CommandRunner` take a `GlobalSession`: I considered this, but it really complicates the whole mechanism.  Activities that access the clients command DB don't really have a GlobalSession, _and I don't want them to fake one_.  If you introduce two types of command runner objects, this gets more complicated still, for very little value.

So I introduced a janky "remove this" token and called it a day.

liuche:
This looks fine to me, too. Adding and storing the unregister-able command seems like acceptable jank to keep the object dependencies relatively clean.

liuche:
refactor looks good to me, r=liuche.

PR #212, opened 2012-06-08
Bug 740193 - Send tab to device.

rnewman:
Obsoletes #161. Needs a little more work to add a migration for command processing.


rnewman:
Note that this subsumes #211.

ncalexan:
Glad this doesn't need re-review, or perhaps only focused review :)  Both this and #211 pass `mvn clean integration-test` (after appropriate preprocessing).

rnewman:
Please check 46bd93f, 1ee2123, 7c4223f, 06e17dd. And you might care about e2e8d98, Nick.

ncalexan:
Everything looks good except the one thing I raised in the diff.  I also fixed a small bug with `GlobalSession.abort()` (see pushed commit).

src/main/java/org/mozilla/gecko/sync/stage/SyncClientsEngineStage.java
```diff
@@ -223,22 +248,35 @@ public String ifUnmodifiedSince() {
...
+      if (!currentlyUploadingLocalRecord) {
+        // TODO: check failed uploads in body.
+        clearRecordsToUpload();
+        checkAndUpload();
         return;
       }
+
+      // If we're processing our record, we have a little more cleanup
+      // to do.
+      commandsProcessedShouldUpload = false;
```
ncalexan, line 278: I think I see what you've changed with the two timestamps.  Is the idea your change will only persist the local client record upload timestamp after all retries and processing are completed?

rnewman, line 278: Well, mostly Marina's code :D

What I changed was:
- No need to look at the body timestamp, because that server bug was fixed.
- Never a need to track per-record timestamps; we just want to know the last time we downloaded or uploaded stuff to the server. Yes, if there's concurrent modification this can lead to us pushing a change to a record that's already been altered, but that's a really narrow window. So always use `ServerClientsTimestamp` for `XIUS`.
- Persist the client record timestamp if we downloaded our record, or uploaded our record (because on upload the header timestamp is the same as the record modified time).

I don't believe this code ever intermingles our record with other clients' records, so the two code paths are separate.

PR #213, opened 2012-06-08
Bug 762978 - Make AndroidBrowserRepositorySession.recordToGuid map from ...

ncalexan:
...hashes of record strings to record guids to reduce memory pressure.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserRepositorySession.java
+++ b/src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserRepositorySession.java
@@ -57,7 +57,21 @@ public abstract class AndroidBrowserRepositorySession extends StoreTrackingRepos
   public static final String LOG_TAG = "BrowserRepoSession";
 
   protected AndroidBrowserRepositoryDataAccessor dbHelper;
-  private HashMap<String, String> recordToGuid;
+
+  /**
+   * In order to reconcile the "same record" with two *different* GUIDs (for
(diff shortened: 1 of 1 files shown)


indygreg:
I thought internally strings were represented as integers when stored in data structures. Or, is it just C# that makes this optimization?

ncalexan:
It's possible, but I've not heard it.  (I'm hardly a JVM expert.)  Even if they are hashed in some way, the Strings must still be in memory, since the key set can be re-constructed.  Here the key set Strings take a lot of memory and I can inexpensively re-create any individual String, so storing only integers should reduce the amount of memory for storing the Map.

src/main/java/org/mozilla/gecko/sync/repositories/android/AndroidBrowserRepositorySession.java
```diff
@@ -660,7 +695,7 @@ private void createRecordToGuidMap() throws NoGuidForIdException, NullCursorExce
         if (record != null) {
           final String recordString = buildRecordString(record);
           if (recordString != null) {
-            recordToGuid.put(recordString, record.guid);
+            recordToGuid.put(recordString.hashCode(), record.guid);
```
liuche, line 698: You use new Integer(hashCode) in other places (I don't think that's actually necessary), but nit for consistency.

liuche:
r+ from me.

ncalexan:
This got merged under the tree name nalexander/bug-749729-shrink-guid-map.

PR #217, opened 2012-06-19
Bug 731116 - More strict mode during setup.

liuche:
Extract calls to AccountManager from main thread.


ncalexan:
Uh, so this is weird: my review commit pushed to the wrong tree.  anyway, look at https://github.com/mozilla-services/android-sync/commit/d7ad59d1f46f

ncalexan:
r=nalexander

PR #219, opened 2012-06-21
Bug 766499 -  Show specific error message when Android fails on Account creation.

liuche:
Added necessary strings. We now show a non-generic error if sync setup fails on Account creation, as well as user-visible reminder that Sync only supports on installed Firefox instance.


liuche:
Updated, with changes and comments added.

PR #220, opened 2012-06-21
Bug 763020 and Bug 763021 - sync subset of engines and do a clients only sync after Send Tab

ncalexan:
I spent a few days implementing a data driven sync, computing dependencies with a topological sort, etc, etc, and when I was done I looked at the patch set and decided that it was a lot of churn for not a lot of win.  Our dependency tree is a line with a diamond of ServerSyncStages (with no interdependencies) in the middle. I think the approach of this ticket, which just has each stage check to see if it is disabled (like we already do if the stage is not present in meta/global) supports this dependency graph well.

See 763021 for detailed device testing instructions.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/GlobalSession.java
+++ b/src/main/java/org/mozilla/gecko/sync/GlobalSession.java
@@ -162,6 +162,11 @@ public class GlobalSession implements CredentialsSource, PrefsSource, HttpRespon
 
     registerCommands();
     prepareStages();
+    Collection<String> knownStageNames = new HashSet<String>();
+    for (Stage stage : Stage.getNamedStages()) {
+      knownStageNames.add(stage.getRepositoryName());
+    }
+    config.stagesToSync = Utils.getStagesToSyncFromBundle(knownStageNames, extras);
 
     // TODO: data-driven plan for the sync, referring to prepareStages.
   }
--- a/src/main/java/org/mozilla/gecko/sync/SyncConfiguration.java
+++ b/src/main/java/org/mozilla/gecko/sync/SyncConfiguration.java
@@ -204,6 +204,9 @@ public class SyncConfiguration implements CredentialsSource {
    * <p>
    * Generated <it>each sync</it> from extras bundle passed to
    * <code>SyncAdapter.onPerformSync</code> and not persisted.
+   * <p>
+   * Not synchronized! Set this exactly once per global session and don't modify
+   * it -- especially not from multiple threads.
    */
   public Collection<String> stagesToSync;
 
--- a/src/main/java/org/mozilla/gecko/sync/SyncConfiguration.java
+++ b/src/main/java/org/mozilla/gecko/sync/SyncConfiguration.java
@@ -6,6 +6,7 @@ package org.mozilla.gecko.sync;
 
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
@@ -195,7 +196,16 @@ public class SyncConfiguration implements CredentialsSource {
    * Copied from latest downloaded meta/global record and used to generate a
    * fresh meta/global record for upload.
    */
-  public Set<String>     enabledEngineNames;
+  public Set<String> enabledEngineNames;
+
+  /**
+   * Names of stages to sync <it>this sync</it>, or <code>null</code> to sync
+   * all known stages.
+   * <p>
+   * Generated <it>each sync</it> from extras bundle passed to
+   * <code>SyncAdapter.onPerformSync</code> and not persisted.
+   */
+  public Collection<String> stagesToSync;
 
   // Fields that maintain a reference to a SharedPreferences instance, used for
   // persistence.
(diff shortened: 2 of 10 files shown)


ncalexan:
rnewman, perhaps you should look at this, since we've discussed this quite a bit.

src/main/java/org/mozilla/gecko/sync/GlobalSession.java
```diff
@@ -162,6 +162,11 @@ public GlobalSession(String userAPI,
 
     registerCommands();
     prepareStages();
+    Collection<String> knownStageNames = new HashSet<String>();
+    for (Stage stage : Stage.getNamedStages()) {
+      knownStageNames.add(stage.getRepositoryName());
+    }
```
rnewman, line 168: Seems kinda redundant to compute this unnecessarily. Why not have `getStagesToSyncFromBundle` return `null` and do this conditionally?

Or, more cleanly, move `getStagesToSync*` into `Stage`, so it knows about the named stages, and you can avoid some of the generality and complexity.

src/main/java/org/mozilla/gecko/sync/SyncConfiguration.java
```diff
@@ -195,7 +196,16 @@ public void unregisterOnSharedPreferenceChangeListener(OnSharedPreferenceChangeL
...
-  public Set<String>     enabledEngineNames;
+  public Set<String> enabledEngineNames;
+
+  /**
+   * Names of stages to sync <it>this sync</it>, or <code>null</code> to sync
+   * all known stages.
+   * <p>
+   * Generated <it>each sync</it> from extras bundle passed to
+   * <code>SyncAdapter.onPerformSync</code> and not persisted.
+   */
```
rnewman: Add a note about not being synchronized…

src/main/java/org/mozilla/gecko/sync/setup/Constants.java
```diff
@@ -18,6 +18,24 @@
...
   public static final String DATA_ENABLE_ON_UPGRADE = "data.enableOnUpgrade";
 
+  /**
+   * Key in sync extras bundle specifying stages to sync this sync session.
+   * <p>
+   * Corresponding value should be a String JSON-encoding an object, the keys of
+   * which are the stage names to sync. For example:
+   * <code>"{ \"stageToSync\": 0 }"</code>.
+   */
+  public static final String EXTRAS_KEY_STAGES_TO_SYNC = "stagesToSync";
```
rnewman: What's wrong with "sync" and "skip"?

src/main/java/org/mozilla/gecko/sync/syncadapter/SyncAdapter.java
```diff
@@ -221,6 +223,29 @@ public boolean shouldBackOff() {
...
+   */
+  public static void requestImmediateSync(final Account account, final String[] stageNames) {
+    if (account == null) {
+      Logger.warn(LOG_TAG, "Not requesting immediate sync because Android Account is null.");
+      return;
+    }
+
+    final Bundle extras = new Bundle();
+    Utils.putStageNamesToSync(extras, stageNames, null);
+    extras.putBoolean(ContentResolver.SYNC_EXTRAS_MANUAL, true);
```
rnewman, line 245: We check for "force" somewhere else in the code (skipping backoff ourselves…?). Clean that up, too?

src/main/java/org/mozilla/gecko/sync/syncadapter/SyncAdapter.java
```diff
@@ -234,6 +259,10 @@ public void onPerformSync(final Account account,
     this.syncResult   = syncResult;
     this.localAccount = account;
 
+    Log.i(LOG_TAG, "Syncing client named " + getClientName() +
+                   " with client guid " + getAccountGUID() +
+                   " (sync account has " + getClientsCount() + " clients).");
+
```
rnewman: Did you leave this here deliberately?

test/src/org/mozilla/gecko/sync/test/TestUtils.java
```diff
@@ -0,0 +1,77 @@
+package org.mozilla.gecko.sync.test;
```
rnewman, line 4: License.

test/src/org/mozilla/gecko/sync/test/TestUtils.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.mozilla.android.sync.test.AndroidSyncTestCase;
+import org.mozilla.gecko.sync.Utils;
+
+import android.os.Bundle;
+
+public class TestUtils extends AndroidSyncTestCase {
```
rnewman, line 18: `TestStageComputation`?

rnewman:
One more go 'round. Looking good.

ncalexan:
liuche:  I renamed `s1` and `s2`.  No tests because I didn't feel like re-working existing code to support testing.  The STRs on the ticket are pretty clear.  If you need to see this tested I'll re-work things.

PR #221, opened 2012-06-21
Bug 741986 - sync scheduling

ncalexan:
In spirit, this is a follow-on to Bugs 763020 and 763021, although I'm not quite there yet.  I'd mostly like feedback on this approach of separating backoffs from regular sync intervals.

I started to complete Bug 741986, but need to discuss what we want to do there.  What I'd like to do is sync clients only perhaps once an hour, as needed, but that will mean introducing another timer pref, etc, and I got to wondering if we want to introduce a timer for every sync stage... we should talk about this.


ncalexan:
We can come back to this when we have time.

PR #223, opened 2012-06-26
Bug 767596 - improve logging

ncalexan:
This basically makes Log.INFO level present the user a very concise "here's what sync did" log.  Would like to do the same for Log.DEBUG level, but not right now.
- Log less PII
- log less at Log.INFO
- more Logger in SyncAdapter


liuche:
Went through more repository-related code, and manual setup authentication.

ncalexan:
Excellent catches!  How do you feel about the log level changing I did, to shorten the log at Log.INFO level?

liuche:
r=liuche

PR #224, opened 2012-06-29
Bug 761682 - version prefs

ncalexan:
First few commits basically stand alone.


rnewman:
Something I didn't see, scanning through these commits: where do you persist which version we have stored in prefs? I'd like to see that explicit, or we're basically guessing which versions from which and to which we're migrating, right?

rnewman:
Looking good.

ncalexan:
Further to your comment about version persistence: none of this code is wired up yet.  Storing the current version will happen outside of the actual prefs file; that's coming shortly.

ncalexan:
Ready for first review, I think.  Patch series:
- versions authtokenType and returns JSON authToken (not turned on yet -- see `SyncAdapter.onPerformSync`, which calls `blockingFromAndroidAccountV0`);
- versions prefs and migrates version 0 forward (and backward);
- writes pickled account prefs to 'sync.account.json', including cleartext username, password, and sync key;
- implements unpickling account prefs.

The tests may or may not play nicely with installed Fennec, but we can continue to work on that.

ncalexan:
This pickles your Sync account settings, and I don't think it needs to.  Will update pull req with fix or justification.

ncalexan:
Closed in favour of

https://github.com/mozilla-services/android-sync/pull/225

PR #226, opened 2012-07-06
Bug 721760 - Correctly open Accounts settings page on Motorola Blur devices.

rnewman:


rnewman:
Note that this should be refactored to do this work in a single place, and lift out constants…

PR #228, opened 2012-07-09
Bug 771024 - Send Tab shows partially truncated list of shareable devices.

liuche:
Scrolling wasn't actually the problem, the list was partially obscured in the layout.

### DIFF
--- a/res/layout/sync_send_tab.xml
+++ b/res/layout/sync_send_tab.xml
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="utf-8"?>
 <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
-  style="@style/SyncLayout" >
+  style="@style/SyncLayout.Vertical" >
   <LinearLayout
     android:id="@+id/sendtab_top"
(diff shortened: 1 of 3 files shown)


ncalexan:
Sorry that this took a day to get to.  On my Samsung Galaxy SII, scrolling and actually selecting check boxes interact badly. I select a box, then tap and drag to scroll, and end up with nothing selected.  I think the original code is really shitty and does click detection itself, rather than using `MultiSelectAndroidBuiltinThing`, but no sense exposing broken scrolling.

liuche:
Tried it out, no more strict mode, doesn't crash, and SendTab scrolling works as expected now.

Thanks for the cleanup and polish, I think this is good to go.

r=liuche

ncalexan:
Further: why two bug numbers?  I can't even access Bug 772789, so I'm going to assume it's wrong.  I copied your bug number (incorrectly) :(

Can you please include bug numbers in your branch names, too?  Essential for keeping my hundreds of trees straight.

I'm going to make all bug numbers 771024 and merge now.

PR #231, opened 2012-07-11
Bug 772978 - log stats

ncalexan:
Really small, but I want some help understanding sync performance differences between my devices.


liuche:
r=liuche, with tiny change in method comment documentation.

PR #232, opened 2012-07-11
Bug 772971 - Pickle Sync account parameters on background thread to avoi...

ncalexan:
...d strict mode warnings.


ncalexan:
I made a mistake separating commits and this doesn't actually compile.  liuche, you need to actually make sure the commit does what it says it will.

PR #234, opened 2012-07-13
Bug 773562 - android.content.ActivityNotFoundException opening Sync sett...

ncalexan:
...ings on Motorola devices.

Oh, Android.  Work around by catching Activity and falling back to
default Sync settings activity.  I re-worked the code because I
couldn't convince myself that intent was always non-null.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/setup/SyncAccounts.java
+++ b/src/main/java/org/mozilla/gecko/sync/setup/SyncAccounts.java
@@ -17,6 +17,7 @@ import org.mozilla.gecko.sync.syncadapter.SyncAdapter;
 
 import android.accounts.Account;
 import android.accounts.AccountManager;
+import android.content.ActivityNotFoundException;
 import android.content.ContentResolver;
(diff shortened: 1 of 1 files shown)


src/main/java/org/mozilla/gecko/sync/setup/SyncAccounts.java
```diff
@@ -364,25 +365,40 @@ public static void setSyncAutomatically(Account account, boolean syncAutomatical
...
+   *
+   * @param context current Android context.
+   * @return the <code>Intent</code> started.
+   */
   public static Intent openSyncSettings(Context context) {
-    Intent intent = null;
-
+    // Bug 721760 - opening sync settings takes user to "Battery & data manager"
+    // on a variety of Motorola devices. This work around tries to hand load the
+    // correct Intent by hand. Oh, Android.
```
rnewman: "hand load by hand".

Sync settings.

Battery & Data Manager

rnewman:
r+ after comment tweaks.

PR #236, opened 2012-07-19
Bug 775360 - "Show me how" link should not launch Browser Intents page.

liuche:
This now launches Fennec instead of the browser choice window.

However, some "back" behavior is weird. If you launch Sync setup from Fennec, hitting back from the "Show me how" link follows the Fennec activity stack, instead of bringing you back to Sync setup. Tried a few flags, didn't seem to fix anything. May be a Fennec problem, because built-in browser handles "back" action correctly, but investigating further.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/setup/activities/WebViewActivity.java
+++ b/src/main/java/org/mozilla/gecko/sync/setup/activities/WebViewActivity.java
@@ -10,6 +10,8 @@ import org.mozilla.gecko.sync.Logger;
 import android.app.Activity;
 import android.net.Uri;
 import android.os.Bundle;
+import android.view.Window;
+import android.webkit.WebChromeClient;
 import android.webkit.WebView;
 import android.webkit.WebViewClient;
 
@@ -24,6 +26,7 @@ public class WebViewActivity extends Activity {
   @Override
   public void onCreate(Bundle savedInstanceState) {
     super.onCreate(savedInstanceState);
+    getWindow().requestFeature(Window.FEATURE_PROGRESS);
     setContentView(R.layout.sync_setup_webview);
     // Extract URI to launch from Intent.
     Uri uri = this.getIntent().getData();
@@ -33,8 +36,16 @@ public class WebViewActivity extends Activity {
     }
 
     WebView wv = (WebView) findViewById(R.id.web_engine);
+    // Add a progress bar.
+    final Activity activity = this;
+    wv.setWebChromeClient(new WebChromeClient() {
+      public void onProgressChanged(WebView view, int progress) {
+        // Activities and WebViews measure progress with different scales.
+        // The progress meter will automatically disappear when we reach 100%
+        activity.setProgress(progress * 100);
+      }
+    });
     wv.setWebViewClient(new WebViewClient() {
-      
       // Handle url loading in this WebView, instead of asking the ActivityManager.
       @Override
       public boolean shouldOverrideUrlLoading(WebView view, String url) {
--- a/src/main/java/org/mozilla/gecko/sync/setup/activities/WebViewActivity.java
+++ b/src/main/java/org/mozilla/gecko/sync/setup/activities/WebViewActivity.java
@@ -14,26 +14,21 @@ import android.webkit.WebView;
 import android.webkit.WebViewClient;
 
 /**
- * Displays Uri in an embedded WebView. Closes if there no Uri is passed in.
+ * Displays URI in an embedded WebView. Closes if there no URI is passed in.
  * @author liuche
  *
  */
 public class WebViewActivity extends Activity {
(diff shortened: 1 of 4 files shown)


ncalexan:
Make sure I understand when this needs review.  (And point to STR on ticket, please!)

rnewman:
Chenxia: any news on this?

liuche:
Accidentally some words. Ready for review!

rnewman:
Did you get the "back" behavior to make sense?

liuche:
Yep, the "show me how" link now displays in a WebView embedded in a new activity, so back behavior is as expected. This activity will show up in "recent activities."

src/main/java/org/mozilla/gecko/sync/setup/activities/WebViewActivity.java
```diff
@@ -0,0 +1,53 @@
...
+import org.mozilla.gecko.sync.Logger;
+
+import android.app.Activity;
+import android.net.Uri;
+import android.os.Bundle;
+import android.webkit.WebView;
+import android.webkit.WebViewClient;
+
+/**
+ * Displays Uri in an embedded WebView. Closes if there no Uri is passed in.
```
rnewman: s/Uri/URI/g

src/main/java/org/mozilla/gecko/sync/setup/activities/WebViewActivity.java
```diff
@@ -0,0 +1,53 @@
...
+ * @author liuche
+ *
+ */
+public class WebViewActivity extends Activity {
+  private final String LOG_TAG = "WebViewActivity";
```
rnewman: `static`.

src/main/java/org/mozilla/gecko/sync/setup/activities/WebViewActivity.java
```diff
@@ -0,0 +1,53 @@
...
+
+  public WebViewActivity() {
+    super();
+    Logger.info(LOG_TAG, "WebViewActivty constructor called.");
```
rnewman: Kill this log line. (And: typo.)

src/main/java/org/mozilla/gecko/sync/setup/activities/WebViewActivity.java
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  @Override
+  public void onCreate(Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+    setContentView(R.layout.sync_setup_webview);
+    // Extract uri to launch from Intent.
+    Uri uri = this.getIntent().getData();
+    if (uri == null) {
+      Logger.debug(LOG_TAG, "No Uri passed to display.");
```
rnewman: s/Uri/URI

rnewman:
Looks good to me. Nits, land (`hg add`!), STR! :D

ncalexan:
Why are we launching a non-Fennec browser at all? This is not smelling right to me.

rnewman:
Firstly, Fennec might not be running, and WebKit is already loaded. Secondly, we don't want users getting into recursive Sync setup, and we can't yet count on Fennec's embeddability.

ncalexan:
On Aug 12, 2012 1:00 PM, "rnewman" notifications@github.com wrote:

> Firstly, Fennec might not be running, and WebKit is already loaded.
> Secondly, we don't want users getting into recursive Sync setup, and we
> can't yet count on Fennec's embeddability.

Why are we embedding at all?  This is exactly what the Android intent
mechanism is for.  If we're trying not to launch Fennec, why write a web
browser at all?  I really feel that this is a bad precedent: Services code
on Android is going to point users to Internet resources and I will fight
to the death to ensure we don't embed WebKit in all those places. If we
can't make Fennec or an intent work, we're doing something gravely wrong.

rnewman:
If we don't embed, we have two choices.

We can fire a general intent, and the first thing the user sees is "which browser do you want to use to open this helpful HTML?", which is not good UX when we've just installed a browser. That's what we used to do, and Bug 775360 was filed as a result.

We can launch Fennec directly — the approach that Chenxia took first — which fucks up our back stack, because it's part of our task (https://bugzilla.mozilla.org/show_bug.cgi?id=775360#c5).

(As a side consequence, we could also get into setup loops, have to wait for the browser to launch if they've come from Settings, etc.)

So we embed, and we have a choice between trying to embed Fennec (which I think it's fair to characterize as "not a ten line patch"), or embedding WebKit.

The latter is readily available, does just fine for displaying a small chunk of help content, and is so easy to do that it's already landed.

I agree with you that eating our own dogfood is a good thing, but pragmatism wins out here: a user clicking "show me how" doesn't care which rendering engine displays that content, and we would do more harm than good by insisting on principle.

(Closing this pull request because this is in m-c; feel free to continue discussion.)

PR #237, opened 2012-07-20
Bug 721760 and Bug 774233: try to load vendor specific Sync settings for...

ncalexan:
... Moto Blur and HTC devices.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/setup/SyncAccounts.java
+++ b/src/main/java/org/mozilla/gecko/sync/setup/SyncAccounts.java
@@ -366,38 +366,68 @@ public class SyncAccounts {
   }
 
   /**
-   * Start sync settings activity.
+   * Bug 721760: try to start a vendor-specific Accounts & Sync activity on Moto
+   * Blur devices.
+   * <p>
+   * Bug 773562: actually start and catch <code>ActivityNotFoundException</code>,
+   * rather than just returning the <code>Intent</code> only, because some
+   * Moto devices fail to start the activity.
    *
-   * @param context current Android context.
+   * @param context
+   *          current Android context.
+   * @param vendorPackage
+   *          vendor specific package name.
+   * @param vendorClass
+   *          vendor specific class name.
+   * @return null on failure, otherwise the <code>Intent</code> started.
+   */
+  protected static Intent openVendorSyncSettings(Context context, final String vendorPackage, final String vendorClass) {
+    try {
+      final int contextFlags = Context.CONTEXT_INCLUDE_CODE | Context.CONTEXT_IGNORE_SECURITY;
+      Context foreignContext = context.createPackageContext(vendorPackage, contextFlags);
+      Class<?> klass = foreignContext.getClassLoader().loadClass(vendorClass);
+
+      final Intent intent = new Intent(foreignContext, klass);
+      context.startActivity(intent);
+      Logger.info(LOG_TAG, "Vendor package " + vendorPackage + " and class " +
+          vendorClass + " found, and activity launched.");
+      return intent;
+    } catch (NameNotFoundException e) {
+      Logger.debug(LOG_TAG, "Vendor package " + vendorPackage + " not found. Skipping.");
+    } catch (ClassNotFoundException e) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/mozilla/gecko/sync/setup/SyncAccounts.java
```diff
@@ -366,38 +366,68 @@ public static void setSyncAutomatically(Account account, boolean syncAutomatical
...
+   * @param context
+   *          current Android context.
+   * @return the <code>Intent</code> started.
+   */
+  public static Intent openSyncSettings(Context context) {
+    // Bug 721760 - opening Sync settings takes user to Battery & Data Manager
+    // on a variety of Motorola devices. This work around tries to load the
+    // correct Intent by hand. Oh, Android.
+    Intent intent;
+    if ((intent = openVendorSyncSettings(context, MOTO_BLUR_PACKAGE, MOTO_BLUR_SETTINGS_ACTIVITY)) != null) {
```
rnewman: Split out the assignment from the conditional for clarity, please.

src/main/java/org/mozilla/gecko/sync/setup/SyncAccounts.java
```diff
@@ -366,38 +366,68 @@ public static void setSyncAutomatically(Account account, boolean syncAutomatical
...
+      return intent;
     }
 
     // Open default Sync settings activity.
-    final Intent intent = new Intent(Settings.ACTION_SYNC_SETTINGS);
-    intent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+    intent = new Intent(Settings.ACTION_SYNC_SETTINGS);
+    // Bug 774233: do not start activity as a new task (second run fails on some HTC devices).
```
rnewman, line 430: Tell me this: why did we open it in a new task in the first place? Presumably that fixed a different bug…

rnewman:
Looks good, modulo one nit and one question.

src/main/java/org/mozilla/gecko/sync/setup/SyncAccounts.java
ncalexan, line 430: I really don't know; reading the Android docs at https://developer.android.com/guide/components/tasks-and-back-stack.html, I have no idea why we wanted a new task.  Kicking this to liuche.

ncalexan:
liuche: can you answer the above question?

liuche:
Originally, when writing the Fennec-back code for opening Fennec after setup (SetupSuccessActivity), I was seeing duplicate activities in the back-stack. I added NEW_TASK to remove duplication*. Didn't think it would cause problems, so I was probably careless in copying it to other places as well.

Went through the flow again, and it turns out it's not necessary to launch Sync Settings from SyncSetupActivity using an intent, so I removed it.

*from the docs for FLAG_ACTIVITY_NEW_TASK:
"When using this flag, if a task is already running for the activity you are now starting, then a new activity will not be started; instead, the current task will simply be brought to the front of the screen with the state it was last in."

ncalexan:
nit: no bug number in commit comment.

nit: I'd like to add docstring to the two click handlers making clear that they are referenced from `android:onClick` in xml (which I hate, btw -- impossible to follow code flow with this style -- but I understand why we do this.)

non-nit: we have a bunch of stuff I don't trust (understand) to do with Android stack management.  In `manifests/SyncAndroidManifest_activities.xml.in`:
- one instance of android:taskAffinity="org.mozilla.gecko.sync.setup"
- two instances of android:clearTaskOnLaunch="true"
- two instances of android:noHistory="true"

Plus code specific hacks.  Do you understand all this and think that it is correct?  I find it bizarre that if we successfully set up, click on Fennec, then click back, we get the home screen.  Why are we setting noHistory at all?

ncalexan:
After Vidyo with liuche, I'm happy with our Android stack management.  I'm going to get this onto m-b.

ncalexan:
I want to keep this minimal, since we're landing on m-b, so I opened a new ticket for removing the new task flag more generally:

https://bugzilla.mozilla.org/show_bug.cgi?id=776659

and I pushed liuche's commit to a new tree:

https://github.com/mozilla-services/android-sync/tree/nalexander/bug-776659-remove-new-task-flag/

liuche, can you land this?

PR #238, opened 2012-07-24
Bug 764867 - match Fennec branding

ncalexan:
This is Bug 742478 done right, with some UI/Style clean up thrown in. `lint android-sync` is much happier now.  This is blocking multiple Fennecs.


rnewman:
r+ once you fix shell portability.

PR #239, opened 2012-07-27
Bug 761682 - version prefs and Bug 777973 - don't always invalidate auth token

ncalexan:
This is getting pretty close to multiple Fennecs!  This rebases the prefs versioning code and throws in an alternative to versioning the auth token type.  Ready for first review, I think.


rnewman:
Feeling pretty awesome about this.

rnewman:
“ (Other authenticator-specific values may be returned.) If an auth token could not be fetched, getResult() throws:

```
AuthenticatorException if the authenticator failed to respond ”
```

rnewman:
This looks good; checking a build now.

PR #240, opened 2012-07-30
Bug 778081 - NPE in SendTab activity

ncalexan:
Fixes the bug by being more careful about null handling, and tries to make SendTab more careful about null URI/title as well.

ADDS A STRING.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/CommandProcessor.java
+++ b/src/main/java/org/mozilla/gecko/sync/CommandProcessor.java
@@ -64,12 +64,22 @@ public class CommandProcessor {
       this.args = args;
     }
 
+    /**
+     * Get list of arguments as strings.  Individual arguments may be null.
+     *
+     * @return list of strings.
+     */
     public synchronized List<String> getArgsList() {
       if (argsList == null) {
         ArrayList<String> argsList = new ArrayList<String>(args.size());
 
         for (int i = 0; i < args.size(); i++) {
-          argsList.add(args.get(i).toString());
+          final Object arg = args.get(i);
+          if (arg == null) {
+            argsList.add(null);
+            continue;
+          }
+          argsList.add(arg.toString());
         }
         this.argsList = argsList;
       }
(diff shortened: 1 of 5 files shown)


src/main/java/org/mozilla/gecko/sync/CommandProcessor.java
```diff
@@ -64,12 +64,22 @@ public Command(String commandType, JSONArray args) {
...
+     *
+     * @return list of strings.
+     */
     public synchronized List<String> getArgsList() {
       if (argsList == null) {
         ArrayList<String> argsList = new ArrayList<String>(args.size());
 
         for (int i = 0; i < args.size(); i++) {
-          argsList.add(args.get(i).toString());
+          Object arg = args.get(i);
```
rnewman: `final`.

src/main/java/org/mozilla/gecko/sync/CommandProcessor.java
```diff
@@ -152,7 +162,11 @@ protected static Command parseCommand(ExtendedJSONObject unparsedCommand) {
 
   @SuppressWarnings("unchecked")
   public void sendURIToClientForDisplay(String uri, String clientID, String title, String sender, Context context) {
-    Logger.info(LOG_TAG, "Sending URI to client: " + uri + " -> " + clientID + " (" + title + ")");
+    if (Logger.LOG_PERSONAL_INFORMATION) {
+      Logger.pii(LOG_TAG, "Sending URI to client: " + uri + " -> " + clientID + " (" + title + ").");
+    } else {
+      Logger.info(LOG_TAG, "Sending URI to client " + clientID + ".");
```
rnewman: Let's always log at `INFO`, with a supplemental PII for URI. Otherwise we'll seem to drop messages when you're building with `L_P_I`.

src/main/java/org/mozilla/gecko/sync/setup/activities/SendTabActivity.java
```diff
@@ -143,19 +154,28 @@ public void run() {
...
-    Toast.makeText(this, R.string.sync_text_tab_sent, Toast.LENGTH_LONG).show();
+  protected void notifyAndFinish(final boolean success) {
+    int textId;
+    if (success) {
+      textId = R.string.sync_text_tab_sent;
+    } else {
+      textId = R.string.sync_text_tab_not_sent;
+    }
+
+    Toast.makeText(this, textId, Toast.LENGTH_LONG).show();
```
rnewman, line 178: So if the send fails, do we retry? And if we do, and it fails again, do we keep showing toasts?

rnewman:
Looks good.

src/main/java/org/mozilla/gecko/sync/setup/activities/SendTabActivity.java
ncalexan, line 178: Nah, this is writing to the local DB only.  As soon as the command hits the DB, we're silent (for good and bad). `notifyAndFinish` is only every called once per SendTab, so I'm not worried about multiple toasts.

ncalexan:
Fixed those two review requests, and replied to toast inquiry.

rnewman:
Go go go.

PR #241, opened 2012-07-31
Bug 741960: Fuzz backoff times up to 25% more to avoid client lock-stepp...

ncalexan:
...ing; agrees with desktop.

This is trivial; let's just do it.


rnewman:
r+

PR #242, opened 2012-07-31
Bug 774300 - Sync authentication errors if passwords contain non-ASCII characters.

rnewman:
We get bytes-as-ASCII from desktop, so we need to decode it into UTF-8, then encode the auth header from UTF-8.


rnewman:
Merged as b0d837a597349c121ce9ada65f7a214ffd48ede4 and 480aaa44f7093cb41b49cd4db198aac6974d8bba.

PR #244, opened 2012-08-17
Bug 780279 - Expose SyncAccounts.requestImmediateSync().

ncalexan:
This is the Sync side of things: exposes a convenience method.  You'll need the Fennec side of things (calling said convenience method) from https://bugzilla.mozilla.org/show_bug.cgi?id=780279.


rnewman:
Part of #246.

PR #246, opened 2012-08-17
Bug 728612 - upload tabs

ncalexan:
This tree is a little messy -- remember to rebase before merging.  Actual user experience sucks rocks due to Fennec not writing tabs to disk frequently, but we'll get this in and then address that.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/repositories/android/FennecTabsRepository.java
+++ b/src/main/java/org/mozilla/gecko/sync/repositories/android/FennecTabsRepository.java
@@ -4,7 +4,11 @@
 
 package org.mozilla.gecko.sync.repositories.android;
 
+import java.util.ArrayList;
+
+import org.json.simple.JSONArray;
 import org.mozilla.gecko.db.BrowserContract;
+import org.mozilla.gecko.db.BrowserContract.Tabs;
 import org.mozilla.gecko.sync.Logger;
 import org.mozilla.gecko.sync.repositories.InactiveSessionException;
 import org.mozilla.gecko.sync.repositories.NoContentProviderException;
@@ -18,14 +22,23 @@ import org.mozilla.gecko.sync.repositories.delegates.RepositorySessionGuidsSince
 import org.mozilla.gecko.sync.repositories.delegates.RepositorySessionWipeDelegate;
 import org.mozilla.gecko.sync.repositories.domain.Record;
 import org.mozilla.gecko.sync.repositories.domain.TabsRecord;
+import org.mozilla.gecko.sync.repositories.domain.TabsRecord.Tab;
 
 import android.content.ContentProviderClient;
 import android.content.ContentValues;
 import android.content.Context;
+import android.database.Cursor;
 import android.net.Uri;
 import android.os.RemoteException;
 
 public class FennecTabsRepository extends Repository {
+  protected final String localClientName;
+  protected final String localClientGuid;
+
+  public FennecTabsRepository(final String localClientName, final String localClientGuid) {
+    this.localClientName = localClientName;
+    this.localClientGuid = localClientGuid;
+  }
 
   /**
    * Note that — unlike most repositories — this will only fetch Fennec's tabs,
@@ -35,15 +48,15 @@ public class FennecTabsRepository extends Repository {
    * unless you use {@link #fetch(String[], RepositorySessionFetchRecordsDelegate)}
    * and specify an explicit GUID.
    */
-  public static class FennecTabsRepositorySession extends RepositorySession {
-
-    private static final String LOG_TAG = "FennecTabsSession";
+  public class FennecTabsRepositorySession extends RepositorySession {
+    protected static final String LOG_TAG = "FennecTabsSession";
 
     private final ContentProviderClient tabsProvider;
     private final ContentProviderClient clientsProvider;
 
-    protected ContentProviderClient getContentProvider(final Context context, final Uri uri) throws NoContentProviderException {
+    protected final RepoUtils.QueryHelper tabsHelper;
 
+    protected ContentProviderClient getContentProvider(final Context context, final Uri uri) throws NoContentProviderException {
       ContentProviderClient client = context.getContentResolver().acquireContentProviderClient(uri);
       if (client == null) {
         throw new NoContentProviderException(uri);
@@ -60,7 +73,6 @@ public class FennecTabsRepository extends Repository {
       } catch (Exception e) {}
     }
 
-
     public FennecTabsRepositorySession(Repository repository, Context context) throws NoContentProviderException {
       super(repository);
       clientsProvider = getContentProvider(context, BrowserContract.Clients.CONTENT_URI);
@@ -74,6 +86,8 @@ public class FennecTabsRepository extends Repository {
         // Oh, Java.
         throw new RuntimeException(e);
       }
+
+      tabsHelper = new RepoUtils.QueryHelper(context, BrowserContract.Tabs.CONTENT_URI, LOG_TAG);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


ncalexan:
Rebased to develop.

src/main/java/org/mozilla/gecko/sync/repositories/android/FennecTabsRepository.java
```diff
@@ -88,37 +102,81 @@ public void finish(final RepositorySessionFinishDelegate delegate) throws Inacti
       super.finish(delegate);
     }
 
+    // Default parameters for local data: local client has null GUID. Override
+    // these to test against non-live data.
+    protected String localClientSelection() {
```
rnewman: `static`.

src/main/java/org/mozilla/gecko/sync/repositories/android/FennecTabsRepository.java
```diff
@@ -88,37 +102,81 @@ public void finish(final RepositorySessionFinishDelegate delegate) throws Inacti
...
+    }
+
     @Override
-    public void guidsSince(long timestamp,
-                           RepositorySessionGuidsSinceDelegate delegate) {
-      // Empty until Bug 730039 lands.
+    public void guidsSince(final long timestamp,
+                           final RepositorySessionGuidsSinceDelegate delegate) {
+      // Bug 783692: Now that Bug 730039 has landed, we could implement this,
+      // but it's not a priority since it's not used (yet).
```
rnewman: While we're here, could we add a log line to remark that we're returning nothing?

src/main/java/org/mozilla/gecko/sync/repositories/android/FennecTabsRepository.java
```diff
@@ -88,37 +102,81 @@ public void finish(final RepositorySessionFinishDelegate delegate) throws Inacti
...
       delegate.onGuidsSinceSucceeded(new String[] {});
```
rnewman: And this should really hit the delegate queue.

src/main/java/org/mozilla/gecko/sync/repositories/android/FennecTabsRepository.java
```diff
@@ -88,37 +102,81 @@ public void finish(final RepositorySessionFinishDelegate delegate) throws Inacti
...
+      final Runnable command = new Runnable() {
+        @Override
+        public void run() {
+          Cursor cursor = null;
+          try {
+            // We fetch all local tabs (since the record must contain them all)
+            // but only process the record if the timestamp is sufficiently
+            // recent.
+            cursor = tabsHelper.safeQuery(tabsProvider, ".fetchSince()", null,
+                localClientSelection(), localClientSelectionArgs(), positionAscending);
```
rnewman: `safeQuery` will throw if the cursor is null, so you can structure this pile of code like this:

```
try {
  final Cursor cursor = tabsHelper.safeQuery(…);
  try {
    // Use cursor; it's never null.
    …
  } finally {
    cursor.close();
  }
} catch (Exception e) {
  delegate.onFetchFailed(e, null);
  return;
}
…
```

rnewman: Evaluate `localClientSelection()` and `localClientSelectionArgs()` outside the `Runnable`; it will help the JIT, and simplifies this bit.

src/main/java/org/mozilla/gecko/sync/repositories/android/FennecTabsRepository.java
```diff
@@ -88,37 +102,81 @@ public void finish(final RepositorySessionFinishDelegate delegate) throws Inacti
...
 
     @Override
-    public void fetch(String[] guids,
-                      RepositorySessionFetchRecordsDelegate delegate) {
-      // Incomplete until Bug 730039 lands.
-      // TODO
+    public void fetch(final String[] guids,
+                      final RepositorySessionFetchRecordsDelegate delegate) {
+      // Bug 783692: Now that Bug 730039 has landed, we could implement this,
+      // but it's not a priority since it's not used (yet).
```
rnewman: Again, log and `delegateQueue`.

src/main/java/org/mozilla/gecko/sync/repositories/android/FennecTabsRepository.java
```diff
@@ -219,4 +277,44 @@ public void createSession(RepositorySessionCreationDelegate delegate,
...
+  public static Tab tabFromCursor(final Cursor cursor) {
+    final String title = RepoUtils.getStringFromCursor(cursor, Tabs.TITLE);
+    final String icon = RepoUtils.getStringFromCursor(cursor, Tabs.FAVICON);
+    final JSONArray history = RepoUtils.getJSONArrayFromCursor(cursor, Tabs.HISTORY);
+    final long lastUsed = RepoUtils.getLongFromCursor(cursor, Tabs.LAST_USED);
+
+    return new Tab(title, icon, history, lastUsed);
+  }
+
+  public static TabsRecord tabsRecordFromCursor(final Cursor cursor, final String clientGuid, final String clientName) {
```
rnewman: Javadoc to note that the caller is responsible for closing the cursor.

src/main/java/org/mozilla/gecko/sync/repositories/android/FennecTabsRepository.java
ncalexan: I don't think we can override if it's static.  Am I right?

src/main/java/org/mozilla/gecko/sync/setup/SyncAccounts.java
```diff
@@ -61,6 +61,56 @@
...
+  public static void requestImmediateSync(final Account account, final String[] stageNames) {
+    if (account == null) {
+      Logger.warn(LOG_TAG, "Not requesting immediate sync because Android Account is null.");
+      return;
+    }
+
+    final Bundle extras = new Bundle();
+    Utils.putStageNamesToSync(extras, stageNames, null);
+    extras.putBoolean(ContentResolver.SYNC_EXTRAS_MANUAL, true);
+    ContentResolver.requestSync(account, BrowserContract.AUTHORITY, extras);
```
rnewman: What does this do if the account is disabled for this app?

test/src/org/mozilla/android/sync/test/FennecTabsRepositorySessionTest.java
```diff
@@ -0,0 +1,200 @@
+package org.mozilla.android.sync.test;
```
rnewman: License.

src/main/java/org/mozilla/gecko/sync/repositories/android/FennecTabsRepository.java
rnewman: > I don't think we can override if it's static. Am I right?
> 
> Yes.

If you don't need it parameterized, though, maybe just compute a value in the constructor and use a final local String member instead of a method? IIRC you can then just set that to a different value in the subclass's constructor.

rnewman:
LGTM after those.

src/main/java/org/mozilla/gecko/sync/repositories/android/FennecTabsRepository.java
ncalexan: I do need it parameterized for testing; it says so in the code comment.  I can't override a final member in an anonymous class (not show stopping, but shitty).  This is a fool's errand -- honestly, overriding the method is the best way to do this.

rnewman: Pah, Java! Alright, let's move on with our lives. 

Might need a @SuppressWarnings to avoid some checkstyle etc. warnings re static.

ncalexan:
Rebased and addressed comments.  I removed Bug 780279 (sync tabs after writing to DB) since, per discussion with rnewman, we'll want to address Fennec writing to disk more frequently, before landing that.  So this is r=rnewman but unfortunately needs to land with Fennec changes to have a decent UX.

PR #248, opened 2012-08-22
Bug 769671 - Don't use IOException(Throwable cause), since it is only av...

ncalexan:
...ailable in API level >= 9.


rnewman:
Heh, I seem to recall us starting off with this idiom! So it goes. r+.

PR #251, opened 2012-08-30
Bug 787249: Fix android-sync integration tests broken by Bug 770785.

ncalexan:
No idea how I managed this, but Bug 770785 breaks integration tests completely.  Here's a re-implemented version.  Bug 770785 has landed on m-i, so I'd like to get this landed ASAP, but I'm not so worried about Nightly because it appears that the actual functionality works fine.


ncalexan:
liuche: please review and test that integration tests work on your devices ASAP.

PR #257, opened 2012-09-13
Bug 753878 engines - second review

ncalexan:
This is not quite a straight rebase -- I rebased everything but also made some small changes.  Comments to come.

### DIFF
--- a/manifests/SyncAndroidManifest_activities.xml.in
+++ b/manifests/SyncAndroidManifest_activities.xml.in
@@ -33,15 +33,30 @@
             android:name="org.mozilla.gecko.sync.setup.activities.RedirectToSetupActivity"
             android:windowSoftInputMode="adjustResize|stateHidden"/>
 
+        <!-- Secondary Sync activities. These depend on other activities for context
+             (display a result, or a next step). Since these don't make sense as stand-alone
+             activities, set excludeFromRecents="true" -->
         <activity
             android:theme="@style/SyncTheme"
             android:noHistory="true"
+            android:excludeFromRecents="true"
             android:name="org.mozilla.gecko.sync.setup.activities.SetupFailureActivity" />
 
         <activity
             android:theme="@style/SyncTheme"
             android:noHistory="true"
+            android:excludeFromRecents="true"
             android:name="org.mozilla.gecko.sync.setup.activities.SetupSuccessActivity" />
+
+        <activity
+            android:theme="@style/SyncTheme"
+            android:excludeFromRecents="true"
+            android:taskAffinity="org.mozilla.gecko.sync.setup"
+            android:name="org.mozilla.gecko.sync.config.activities.SelectEnginesActivity">
+            <intent-filter>
+                <action android:name="android.intent.action.MAIN" />
+            </intent-filter>
+         </activity>
 
         <receiver
             android:name="org.mozilla.gecko.sync.receivers.UpgradeReceiver">
@@ -62,6 +77,7 @@
 
         <activity
             android:theme="@style/SyncTheme"
+            android:excludeFromRecents="true"
             android:icon="@drawable/icon"
             android:label="@string/sync_app_name"
             android:configChanges="orientation"
--- a/src/main/java/org/mozilla/gecko/sync/ExtendedJSONObject.java
+++ b/src/main/java/org/mozilla/gecko/sync/ExtendedJSONObject.java
@@ -125,6 +125,7 @@ public class ExtendedJSONObject {
 
   /**
    * Return a server timestamp value as milliseconds since epoch.
+   *
    * @param key
    * @return A Long, or null if the value is non-numeric or doesn't exist.
    */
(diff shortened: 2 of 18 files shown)


ncalexan:
I have some specific comments and some general comments.  Specific comments first:

Fold the two Activities into one, in preparation for making one ConfigureSyncActivity that will be called from Fennec.  Think about the lifecycle of this Activity -- it's not clear to me that onResume is the correct place to configure the Account.  I think the user should be able to go away and come back to configuration.

ncalexan:
General comments: no tests is not awesome-sauce.
- Integration tests for SyncConfiguration.blockingPrefs..., plz.
- Unit tests (to dummy shared prefs) for storing/reading selected engines to/from prefs, demonstrating handling of forms/history.
- Some tests explaining the gnarly userSelected/enabled/default logic would be gravy.

ncalexan:
General comment:

Prefer once and only once, as in http://c2.com/xp/OnceAndOnlyOnce.html.

For example, in SyncAccounts you added two duplicate helper functions.  Use the more general one in the less general one.

Also, you added some nice general functions: blockingPrefs..., validEngineNames.  You even copy-pasta'd one of them!  You need to check where your new function is applicable and try to update the relevant places to use them.

ncalexan:
General comment: Bug XXX at beginning of every commit message.

ncalexan:
General comment: this is looking Pretty Damn Awesome.  Good work!

ncalexan:
Re: updating config: I'm convinced.  I'll try to get to this tomorrow AM; I'm too busy plugging on another bug to context switch this back in.

ncalexan:
Okay, this is looking really close.  I repeat:
- Integration tests for SyncConfiguration.blockingPrefs..., plz.
- Unit tests (to dummy shared prefs) for storing/reading selected engines to/from prefs, demonstrating handling of forms/history.

Then rebase this down to 5 independent patches, like we had before:
- Bug 753878 - Upload meta/global record with user selected engines to sync.
- Bug 753878 - Add activity for user selecting engines to sync.
- Bug 753878 - Add SyncAccounts.blockingPrefsFromAndroidAccountV0.
- Bug 753878 - Add ExtendedJSONObject.remove.
- Bug 753878 - Factor out SyncConfiguration.getEnabledEngineNames.

And we're good to go to QA.

ncalexan:
Ooh, this is looking _excellent_.  Device testing works like a charm for me!

I have only one concern, and I'm not sure what we should do with it.  When you fire up the "Sync my..." dialog, make your selections, and hit "OK", you get a toast saying "Your preferences have been saved."  I don't think our users have the mental model of local preferences/remote preferences and that this will be confusing.

I wonder if we should prompt for a sync, either in the toast ("[Force] Sync now to update your preferences on all your devices.") or just by spinning one up (see `SyncAdapter.requestImmediateSync`).  rnewman?

r=nalexander after we address this.

liuche:
I like the idea of requesting an immediate sync - this is what desktop does, right, instant sync? I can't think of any downsides to requesting immediate syncs, as long as Android can handle things like already syncing or sync disabled - anything come to mind (rnewman, nalexander)?

liuche:
Rebased and added immediate syncing on saving selections.

ncalexan:
I left this issue (requesting sync) open for discussion because it's more complicated than just requesting a sync.  What happens if the sync fails -- does this code handle not uploading meta/global and trying again the next sync?  (I think it will not.  We "complete" a sync even if the net connection is dropped after info/collections, which means `UploadMetaGlobalStage` executes... perhaps this pull is clever enough to handle this.)   Also, is our messaging good in that situation?  (I think not -- it's basically not possible to feed information about a triggered sync back to the shooter.)

liuche:
nalexander, correct on the null argument for requestImmediateSync, fixed now.

Added an incremental clearing of selectedEngines, in case an engine stage doesn't run at all. Review plz, and I will rebase it back into the stack of commits. This should make partially failed syncs more robust.

In the case that meta/global fails to upload, selected engines will be tried on subsequent syncs until a success, when we will clear the updated engines from the selected engines pref.

liuche:
New changes: commits following Part 4 (that should ultimately be folded into part 4).

These clean up engine upload behavior in the cases of 1) incomplete engines sync and 2) timestamp checking for local/remote meta/global to determine whether or not to upload.

liuche:
Rebased on develop, added new test for syncing with an unknown engine, debug level logging for timestamp reconciliation (for observing clock skew between devices), and addressed review comments.

ncalexan:
The test for making sure we don't remove a `new-engine` looks good.  However, I've forgotten what the consensus was around engine uploading with timestamps.  liuche, can you remind me?

ncalexan:
Further to discussion in MV with liuche and rnewman: if two offline clients conflict, then first client to sync wins.  Further to discussion with liuche: r=nalexander after folding commits.  Any additional changes will be follow-up.

liuche:
Merged into develop.

PR #258, opened 2012-09-17
Nalexander/s20

ncalexan:


ncalexan:
Comments noted, but this will be broken into smaller pieces and new PRs opened.

PR #259, opened 2012-09-18
Bug 792254 - move unit tests requiring network connection to integration tests

ncalexan:
This was motivated by really slow responses from remote Sync servers, but will improve our testing in the Jenkins environment as well.

Looking forward to testing more against local servers!


rnewman:
Works for me.

PR #260, opened 2012-09-20
Bug 790931 - remove account

ncalexan:
This turned out to be a bit of a rabbit hole.  The commit messages are well worth the read, I think :) rnewman, liuche: you should both look at this one.

### DIFF
--- a/android-sync-instrumentation/pom.xml
+++ b/android-sync-instrumentation/pom.xml
@@ -72,31 +72,6 @@
     <sourceDirectory>../test/src</sourceDirectory>
 
     <plugins>
-      <!--
-        For some reason, when compiling under Jenkins the compiler
-        doesn't see some of the main app's compiled classes.
-        Fix this by adding part of its source tree to the build.
-        -->
-      <plugin>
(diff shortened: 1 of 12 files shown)


ncalexan:
rnewman: issues addressed and re-pushed :)  Comments gone :(

I blacklisted the Sync key (replaced with empty string, needed because SyncAccountParameters checks for non-null Sync key).  The tests do not actually test invalidating.  They do test fetching an auth token, which is no longer used.  If you want me to remove that, and make SyncAuthenticator throw if you ask for an auth token, we could do that, but let's open a new ticket for that.

rnewman:
I'm happy with this. Please test and make sure it avoids the problem, firstly, and secondly that it's harmless when "half-landed". Then we can land it!

ncalexan:
rnewman: a typo or two crept into `GlobalConstants.java.in`; the recent pushes fix that (and only that).  This has moved to r? on https://bugzilla.mozilla.org/show_bug.cgi?id=790931; let's do the final tracking there.  (I'll still merge here, of course, when it is appropriate.)

PR #261, opened 2012-10-10
Inconsistency between javadoc param and method arg.

liuche:
Missed a javadoc param while refactoring. (engine selection for bug 753878)


rnewman:
Include bug number in commit message, merge this in the same patch that you land on inbound, if you haven't already. r+.

liuche:
I already pushed the engines patch to m-i yesterday, so I'll land this once I get confirmation that that patch made it into the tree.

ncalexan:
rnewman: can you fold this follow up into your fixes for Announcements?  I think that will be next to land.

rnewman:
Done. Closing this.

PR #262, opened 2012-10-13
Bug 801225, Bug 793053 post-review

rnewman:


rnewman:
Dammit, GitHub, show my commits in the right order!

rnewman:
Note that the per-thread logging code hasn't been tested yet.

rnewman:
Note also that I'm leaving GlobalConstants in the org.mozilla.gecko.sync package until we address some more reorganization on the Fennec side. It's not important enough to do now; bigger fish to fry.

rnewman:
Verified that both Sync setup and the new announcements code can run in the same process with different log tags. And If-Unmodified-Since seems to work!

All Sync tests (unit and integration) pass.

rnewman:
Correct order for the cleanup commits:

```
f593e8b Bug 801225 - Part 1: Java hygiene.
53e173a Bug 801225 - Part 2: adjust logging to use ThreadLocals for global tag.
fe89374 Bug 801225 - Part 3: separate GlobalConstants and SyncConstants.
de4073a Bug 801225 - Part 4: clean up imports and logging.
70daf64 Bug 801225 - Part 5: introduce ActivityUtils and SyncActivity to control Sync logging.
2f6a4a4 Bug 801225 - Part 6: use ActivityUtils and SyncActivity.
```

rnewman:
Oh, with

```
 31ec12f Eliminate enablePlainHTTPConnectionManager from tests and BaseResource, after 944d2cd made it obsolete.
```

coming first.

ncalexan:
I can't comment on remove `enablePlainHTTPConnectionManager` because I've never used it and it seems to have no effect.  Certainly test code can access http:// and https:// without special manipulations that I can see.

I have one (one!) question: why `SimpleLogTagWriter`?  Without `ThreadLocalLogTagWriter`, `simple` makes little sense.  I preferred `single`, but would take `unthreaded`.  Otherwise, this looks okay.

Also, have a test! r=nalexander

ncalexan:
Argh: please cherry-pick test from https://github.com/mozilla-services/android-sync/tree/nalexander/announce

rnewman:
I used "Simple" because it takes a string as the log tag, doesn't vary it, and does what you would expect. It's the simplest possible `LogTagWriter`, if you will. `Unthreaded` would imply some kind of concurrency consequences, which I think is misleading.

PR #263, opened 2012-10-16
Bug 802284, Bug 718057 - parameterize test port

ncalexan:
For a few weeks now, we've had intermittent Jenkins failures.  I believe they are caused by resource contention beween the `sync-android` and `sync-android-retest` branches: both test suites bind to port 15325, and the resulting races lead to BindExceptions.  This PR address them.

It protects against same-process resource contention by only allowing one `HTTPServerTestHelper` to be active at once, and throwing helpful stack traces telling you where an existing server was created.

It protects against inter-process resource contention by adding a system property for specifying the test port that can be easily configured in Jenkins.

I have pushed this to `sync-android-retest`.  The following unsuccessful builds used the same port:

https://ci.mozilla.org/job/sync-android/651/
https://ci.mozilla.org/job/sync-android-retest/88/

The following successful builds used different ports:

https://ci.mozilla.org/job/sync-android/651/
https://ci.mozilla.org/job/sync-android-retest/88/

I have now configured the two Jenkins jobs to _not_ conflict, and set them to build every hour.  This should give plenty of data to verify that this is indeed the issue, and that this PR fixes it.  Let's land this ASAP.

### DIFF
--- a/src/test/java/org/mozilla/android/sync/test/helpers/HTTPServerTestHelper.java
+++ b/src/test/java/org/mozilla/android/sync/test/helpers/HTTPServerTestHelper.java
@@ -25,26 +25,84 @@ import org.simpleframework.transport.connect.SocketConnection;
  * <b>unchecked</b> exception thrown contains a stack trace pointing to where
  * the new server is being created and where the pre-existing server was
  * created.
+ * <p>
+ * Parses a system property to determine current test port, which is fixed for
+ * the duration of a test execution.
  */
 public class HTTPServerTestHelper {
   private static final String LOG_TAG = "HTTPServerTestHelper";
+
+  /**
+   * Port to run HTTP servers on during this test execution.
+   * <p>
+   * Lazily initialized on first call to {@link #getTestPort}.
+   */
+  public static Integer testPort = null;
+
+  public static final String LOCAL_HTTP_PORT_PROPERTY = "android.sync.local.http.port";
+  public static final int    LOCAL_HTTP_PORT_DEFAULT = 15125;
 
   public final int port;
 
   public Connection connection;
   public MockServer server;
 
+  /**
+   * Create a helper to bind <code>MockServer</code> instances.
+   * <p>
+   * Use {@link #getTestPort} to determine the port this helper will bind to.
+   */
   public HTTPServerTestHelper() {
     this.port = getTestPort();
-  }
-
-  public synchronized static int getTestPort() {
-    return 5000;
   }
 
   // For testing only.
   protected HTTPServerTestHelper(int port) {
     this.port = port;
+  }
+
+  /**
+   * Lazily initialize test port for this test execution.
+   * <p>
+   * Only called from {@link #getTestPort}.
+   * <p>
+   * If the test port has not been determined, we try to parse it from a system
+   * property; if that fails, we return the default test port.
+   */
+  protected synchronized static void ensureTestPort() {
+    if (testPort != null) {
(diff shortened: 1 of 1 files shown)


src/test/java/org/mozilla/android/sync/test/helpers/HTTPServerTestHelper.java
```diff
@@ -15,52 +17,180 @@
...
+        Logger.warn(LOG_TAG, "Got exception parsing local test port; ignoring. ", e);
+      }
+    }
+
+    if (testPort == null) {
+      testPort = Integer.valueOf(LOCAL_HTTP_PORT_DEFAULT);
+    }
+  }
+
+  public synchronized static int getTestPort() {
```
rnewman, line 100: Doc, please. Does this return the same value if called multiple times within the same thread? Do we need to clean up in some way?

src/test/java/org/mozilla/android/sync/net/test/TestResource.java
```diff
@@ -23,10 +23,10 @@
 import ch.boye.httpclientandroidlib.HttpResponse;
 
 public class TestResource {
-  private static final int    TEST_PORT   = 15325;
+  private static final int    TEST_PORT   = HTTPServerTestHelper.getTestPort();
```
rnewman, line 26: Does this value ever change within an execution context? If yes, then change the name to not suggest `CONSTANT`. If not, then how about we kill this and just allow callers to ask the `HTTPServerTestHelper` (`data`) for its port?

test/src/org/mozilla/android/sync/stage/test/TestClientsStage.java
```diff
@@ -19,8 +19,7 @@
 import android.content.Context;
 
 public class TestClientsStage extends AndroidSyncTestCase {
-  private static int          TEST_PORT        = 15325;
-  private static final String TEST_CLUSTER_URL = "http://localhost:" + TEST_PORT;
+  private static final String TEST_CLUSTER_URL = "http://not.used";
```
rnewman: Not used => kill it.

src/test/java/org/mozilla/android/sync/test/helpers/HTTPServerTestHelper.java
ncalexan, line 100: This is really simplistic: it's just statically cached, so no thread fanciness at all.  Happy to document that fact.  No clean-up needed.

src/test/java/org/mozilla/android/sync/net/test/TestResource.java
ncalexan, line 26: It doesn't.  I started down the "ask the server helper for its port" path, and the tendrils run deep enough that it's not worth fixing.  (A lot of the test code is super brittle :()

test/src/org/mozilla/android/sync/stage/test/TestClientsStage.java
ncalexan: So, it's not used, but it is passed to the `GlobalSession` constructor.  I checked -- we can pass `null` -- but I'm not sure that's a good habit to get into.  I could go either way, since it's only in a couple of places.

ncalexan:
Updated with some javadocs and some URLs removed.  I don't think the more substantive change you are suggesting is a good use of our time, so ask for re-review without it.

rnewman:
Land it!

PR #264, opened 2012-10-20
Nalexander/bug 799734 bid library

ncalexan:
rnewman, how do you feel about reviewing the auth header changes I made to support Basic-Auth, Browser-ID, and MAC?  The first few commits for Bug 799729, that is.


ncalexan:
This has been a clearing house for scraps of code for a while now, so I'm going to remove it from the open pulls list.

PR #266, opened 2012-10-30
Bug 806634 - Make AnnouncementsBroadcastService.recordLastLaunch public.

rnewman:
This is so Fennec can write our startup time.


liuche:
r+ after nit and comments.

PR #268, opened 2012-11-13
Bug 810988 - Part 1: announcements handler changes for broadcast split.

rnewman:
Prefix intent action. Touches manifest fragment and constants file, which is now preprocessed.


ncalexan:
r=nalexander is on the ticket, but let's be thorough.

ncalexan:
Actually, I'd like to request that you add a comment to `AndroidManifest.xml.in` explaining why the Announcements activities, services, etc are not included in `android_sync.apk`.  (Something about requiring Fennec or similar is enough.)

PR #269, opened 2012-11-13
Bug 811508 and Bug 811510 - vagrant configurations for develop build environment and sync20 tokenserver/server 2.0

ncalexan:
This should be merged if `vagrant up` is successful -- there's no harm to be had here.


ncalexan:
This is low impact but I'd like to get these out of my mental queue.  Also, running local Sync server 2.0 is good for testing!

rnewman:
Fine by me!

PR #270, opened 2012-11-14
Bug 799729 - generalize base resource authentication

ncalexan:
rnewman, this one's on you.  I have additional code on top of this that uses the new `AuthHeaderProvider` concrete classes, so I'm confident the testing given is sufficient (even though it's not _live_ testing).


ncalexan:
Aside: cleaning up crufty `BaseResourceDelegate` hierarchy wasn't strictly necessary, but we need to start burying some of our bodies...

PR #273, opened 2012-12-03
Bug 814331 - Use info/collection_counts to avoid hitting bookmark limits.

rnewman:
Needs an integration test, but otherwise I think this is sound. Splits `InfoCollections` in two, introduces a safe kind of repository session and an additional way to short-circuit a stage.


ncalexan:
You beat me to it -- I did half of this last week.  Will review ASAP.

ncalexan:
This code does what it says, but I don't agree with the fundamental premise that the "record type" knows how to fetch.  Separate these functions, and I'm happy.

rnewman:
Yeah, I concur. Was taking the easier route of following existing pattern, largely due to urgency. This is about as compact as I could make this change.

rnewman:
I still want an integration test.

ncalexan:
I'm not sure this needs an integration test: what would it add?  You've hit the main points; anything else will shake out in testing.

r=nalexander after addressing final comments.

rnewman:
Merged… with another test, too.

PR #274, opened 2012-12-04
Bug 817842 — Modify Announcements client to handle clock skew for If-Modified-Since header

rnewman:


ncalexan:
overall: it's good, but I'd like clarity in names and (especially in) comments differentiating last local fetch time and last remote server date.

PR #275, opened 2012-12-04
Bug 818218 - refactor extended json object parsing

ncalexan:
A long overdue cleanup, I think, with the benefit of using a synchronized static single JSONParser instance.

I device tested this to make sure that pairing still works with the parsing changes.


ncalexan:
Okay, more tests, and emphasized the streaming interface more.  It's different enough that it probably all needs re-review.

ncalexan:
For the hell of it, I made `RepositorySessionBundle` not a huge fail.

rnewman:
LGTM!

ncalexan:
rnewman: here's a slightly different end to this sequence.  This just refactors `ExtendedJSONObject` and friends, but doesn't change the underlying `JSONParser` instance creation.  I'd like to land these refactorings, and then consider more fully using Jackson.  Plan?

rnewman:
Sounds good to me.

PR #276, opened 2012-12-04
Bug 815773 - process local commands

ncalexan:
This is just the base stuff to process local commands.  (local commands == commands generated by local actions, such as from Sync > Options).

There's UI work to be done on top of this, but we may as well look over this first.

The key point is that local commands are processed when the clients engine downloads its local GUID; that means there's a race if you haven't yet synced.

Braindump 1: there could be interactions if you reset your local GUID that I haven't considered.
Braindump 2: I should verify that local commands are deleted from the DB after processing.  I think I did this, but I should verify.  This probably needs to wait until UI is in place.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/CommandProcessor.java
+++ b/src/main/java/org/mozilla/gecko/sync/CommandProcessor.java
@@ -5,6 +5,7 @@
 package org.mozilla.gecko.sync;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
@@ -39,11 +40,12 @@ import android.net.Uri;
  * long-lived singleton.
  */
 public class CommandProcessor {
-  private static final String LOG_TAG = "Command";
+  private static final String LOG_TAG = CommandProcessor.class.getSimpleName();
+
   private static AtomicInteger currentId = new AtomicInteger();
   protected ConcurrentHashMap<String, CommandRunner> commands = new ConcurrentHashMap<String, CommandRunner>();
 
-  private final static CommandProcessor processor = new CommandProcessor();
+  private static CommandProcessor processor = null;
 
   /**
    * Get the global singleton command processor.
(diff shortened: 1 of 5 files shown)


src/main/java/org/mozilla/gecko/sync/CommandProcessor.java
```diff
@@ -39,11 +40,12 @@
...
  */
 public class CommandProcessor {
-  private static final String LOG_TAG = "Command";
+  private static final String LOG_TAG = CommandProcessor.class.getSimpleName();
+
   private static AtomicInteger currentId = new AtomicInteger();
   protected ConcurrentHashMap<String, CommandRunner> commands = new ConcurrentHashMap<String, CommandRunner>();
 
-  private final static CommandProcessor processor = new CommandProcessor();
+  private static CommandProcessor processor = null;
```
rnewman, line 49: Making this non-final means that it's possible for registered commands to get lost. You need to define under what circumstances this processor is released, and when commands should register with a new one.

src/main/java/org/mozilla/gecko/sync/CommandProcessor.java
```diff
@@ -51,9 +53,64 @@
    * @return the singleton processor.
    */
   public static CommandProcessor getProcessor() {
+    if (processor != null) {
+      return processor;
+    }
```
rnewman, line 58: Access to this method needs to be synchronized.

src/main/java/org/mozilla/gecko/sync/stage/SyncClientsEngineStage.java
```diff
@@ -87,6 +89,14 @@ protected synchronized void closeDataAccessor() {
     db = null;
   }
 
+  protected synchronized CommandProcessor getCommandProcessor() {
```
rnewman, line 92: Don't do this. Call `CommandProcessor.getProcessor()` directly — otherwise you could violate singletonness.

src/main/java/org/mozilla/gecko/sync/stage/SyncClientsEngineStage.java
```diff
@@ -405,7 +420,7 @@ protected void processCommands(JSONArray commands) {
     }
 
     commandsProcessedShouldUpload = true;
-    CommandProcessor processor = CommandProcessor.getProcessor();
+    CommandProcessor processor = getCommandProcessor();
```
rnewman, line 423: Call directly every time you need to use it. Because the `CommandProcessor` singleton is no longer `final`, it's not safe to cache this return value anywhere.

src/main/java/org/mozilla/gecko/sync/stage/SyncClientsEngineStage.java
rnewman, line 92: Oh, you're doing this for testing. In that case, make `getCommandProcessor` a wrapper around `CommandProcessor.getProcessor` — don't cache at all.

src/main/java/org/mozilla/gecko/sync/stage/SyncClientsEngineStage.java
```diff
@@ -196,6 +206,11 @@ public void handleWBO(CryptoRecord record) {
           localAccountGUIDDownloaded = true;
           session.config.persistServerClientRecordTimestamp(r.lastModified);
           processCommands(r.commands);
+
+          // It is possible that there are commands in the database for the
+          // local GUID that did not come from the server, namely locally
+          // triggered "wipeAll", "resetAll", etc.
+          processLocalCommands(r.guid);
```
rnewman, line 213: Can you justify why you run server commands then local commands?

Would it be better the other way 'round?

Would it be better to process the two lists merged together and sorted by date?

rnewman, line 213: I think you might be doing this in the wrong place.

You're only running local commands when we process an incoming record for our GUID — e.g., if someone sends us a command. Shouldn't we run local commands every time we sync?

rnewman:
Do you have a new version of this, Nick?

ncalexan:
No updated version yet.  Your suggestions are good but make this significantly more work ;)

PR #277, opened 2013-01-02
Bug 799729 - Let HMAC extra parameter be the empty string.

ncalexan:
This is against the spec (!), but seems natural to me.  It is assumed
by existing Mozilla Services server implementations, including the
token server.


rnewman:
Fine by me, but please add a comment that this is against the spec. And ideally fix the spec ;)

PR #280, opened 2013-01-12
Bug 799615 - send tab from Twitter shares

ncalexan:
This is a heuristic, and it might not be the right heuristic, but it's simple.  r=, rnewman?

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/setup/activities/SendTabData.java
+++ b/src/main/java/org/mozilla/gecko/sync/setup/activities/SendTabData.java
@@ -1,0 +1,61 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.gecko.sync.setup.activities;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import android.content.Intent;
+import android.os.Bundle;
+
+/**
+ * A static factory that extracts (title, uri) pairs suitable for send tab from
+ * Android intent instances.
+ * <p>
+ * Takes some care to extract likely "Web URLs" in preference to general URIs.
+ */
+public class SendTabData {
+  public final String title;
+  public final String uri;
+
+  public SendTabData(String title, String uri) {
+    this.title = title;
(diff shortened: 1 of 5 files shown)


rnewman:
Furthermore: maybe even `findFirstWebURL`. We absolutely do not want to allow a `javascript:` or `file:` URI to be sent, for example. So let's take a test for that.

Otherwise, looks good!

ncalexan:
Reworked this to have a stand-alone WebURLFinder.  Re-review, please.

src/main/java/org/mozilla/gecko/sync/setup/activities/SendTabData.java
```diff
@@ -0,0 +1,62 @@
...
+      StringBuffer sb = new StringBuffer();
+      for (String data : new String[] { text, subject, title }) {
+        if (data != null) {
+          sb.append(data);
+          sb.append("\n\n");
+        }
+      }
+
+      // Find first URL.
+      theUri = new WebURLFinder(sb.toString()).bestWebURL();
```
rnewman: Why not just have WebURLFinder take a collection? Cheaper than concatenating a new string.

src/main/java/org/mozilla/gecko/sync/setup/activities/SendTabData.java
```diff
@@ -0,0 +1,55 @@
...
+
+package org.mozilla.gecko.sync.setup.activities;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import android.content.Intent;
+import android.os.Bundle;
+
+public class SendTabData {
```
rnewman: Small comment.

src/main/java/org/mozilla/gecko/sync/setup/activities/WebURLFinder.java
```diff
@@ -0,0 +1,137 @@
...
+
+public class WebURLFinder {
+  public final List<String> candidates;
+
+  public WebURLFinder(String string) {
+    if (string == null) {
+      throw new IllegalArgumentException("string must not be null");
+    }
+
+    List<String> strings = new ArrayList<String>();
```
rnewman: But there's a `candidateWebURLs` method that takes a single string! Just use that!

rnewman:
r+

PR #281, opened 2013-01-16
Bug 815838 - developer tools

ncalexan:
The dump database activity is very much WIP.  I just was using it to work on the notifications stuff for a different bug.  But the other stuff is pretty small, developer only, and rather useful; good to get merged.


rnewman:
Update needed before review!

PR #282, opened 2013-01-17
Bug 796187 - Send tab: usability tweaks

rnewman:
I ended up reworking the `ClientRecordArrayAdapter`, because despite being an `ArrayAdapter` it elected to manage its own array!

We now populate and connect it, then fetch client records and update the adapter accordingly.

We only fetch client records once, regardless of how many times you resume the activity. That should stop boxes getting unchecked.

And then the real work: check automatically if there's only one client to which to send, and make the rows slightly fatter.

### DIFF
--- a/res/layout/sync_list_item.xml
+++ b/res/layout/sync_list_item.xml
@@ -4,10 +4,11 @@
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
-    android:orientation="horizontal"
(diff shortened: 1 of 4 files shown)


ncalexan:
How do you feel about displaying title and URL?  I like it.

ncalexan:
I have one substantive complaint, and one possible enhancement, but everything else looks good.

ncalexan:
Looks good to me.  If you are happy with displaying the title and uri, rebase this down and r=nalexander.

rnewman:
Merged.

PR #284, opened 2013-01-21
Bug 832773 - Fix Send Tab button enabling logic.

ncalexan:
Well, that is embarrassing.  Should have caught that in review.  This fixes the logic and removes the num checked internal state for good measure.


rnewman:
Weird. r+

PR #285, opened 2013-01-21
Bug 832773 - Follow-up: fix Send Tab button enabling logic on activity resume.

ncalexan:
I missed a case: we weren't updating the send button status on activity resume.  What is really going on is that our life cycle is not correct: we are doing things in `onResume` that are one-time only and should be in `onCreate`.  Our guards were sufficient for some things, but not others -- in particular, the enable state would be hard-set to disabled in `onResume`, when it should not have been touched.


rnewman:
LGTM.

PR #286, opened 2013-01-22
Bug 833464 - set backoff on forced sync

ncalexan:
So, this is a little controversial.  There is a little tidbit, fixing a bad timer, that I'd like to land no matter what.  See ticket comment and commit comment.  I could be convinced that this is a bad idea.


rnewman:
This all looks good to me.

PR #287, opened 2013-01-22
Bug 833508 - two phased global sync stage constructors

ncalexan:
rnewman: we talked about this in the context of data driven global sessions.  This is just a carefully segmented implementation of the two phased construction process that lets me create a `GlobalSyncStage` without tying it immediately to a `GlobalSession`.


rnewman:
I'm happy with this as a starting point. r+

I definitely want to see us move away from `AbstractSessionManagingSyncStage`: it defeats much of the state removal that you're doing here. Possible alternatives:

• Setter for `this.session` that throws if it changes between calls;
• A custom `Runnable` that closes over a session, or similar runtime approach;
• More invasive changes that eliminate use of `this.session` altogether.

But in a way, what we've done here is implement Android's `ApplicationContext`. Interesting to think about.

ncalexan:
Re: removing `AbstractSessionManagingSyncStage`:

You have interpreted this series as commits as intended: they are work in progress to something with less state at construction time.  I actually implemented a "non-final final" with a setter and getter pair, but it's just not my personality.  Not sure how a custom `Runnable` helps here -- that is equivalent to having a `GlobalSyncStage` factory that builds stages parameterized by global sessions at creation time (which is a valid approach, but much heavier than this one).  I prefer more invasive changes that eliminate the session member variable because this will highlight the state going into and out of a sync stage.

I have a pipe dream of untying this from Android entirely and running syncs in a Plain Old Java environment, so `ApplicationContext` is a non-starter for me.  Thanks for review!

rnewman:
By `Runnable` I mean that whatever code that runs asynchronously within one of the methods on the stage — e.g., `resetLocal` — can close over the session to which it should refer. Basically a technique to eliminate `this.session` within a particular scope.

Carry on!

PR #289, opened 2013-01-29
Bug 835992 - format makefile variables

ncalexan:


rnewman:
If it builds, go for it! :D

PR #291, opened 2013-01-31
Bug 836780 - data available

ncalexan:
This might not be our long term strategy, but let's minimize our HTTP GETs with the architecture we have in place.

Those rebuilds?  So I can device test!

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/Logger.java
+++ b/src/main/java/org/mozilla/gecko/sync/Logger.java
@@ -4,6 +4,7 @@
 
 package org.mozilla.gecko.sync;
 
+import java.io.PrintWriter;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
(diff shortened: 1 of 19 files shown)


rnewman:
Can't argue with this!

ncalexan:
This tree actually works on device!  It's still clown shoes all the way: due to the way `begin` and `unbundle` work, we do a ton of DB crawling in bookmarks and history that might not be necessary.  But this at least reduces network requests.

ncalexan:
rnewman: can I get re-review on this?  There were significant changes due to the way timestamps were actually handled.

rnewman:
This looks fine to me. Needs rebasing on top of logger reorg, though.

ncalexan:
Closing in favour of the rebased, updated, and re-tested https://github.com/mozilla-services/android-sync/pull/437.

PR #292, opened 2013-02-06
 Bug 838416 - Defend against insane clocks.

rnewman:
I haven't manually tested this, but it does now have some tests…


rnewman:
QAed and landed.

PR #294, opened 2013-02-19
Bug 842674 - Use correct constant for android.intent.action.EXTERNAL_APPLICATIONS_AVAILABLE.

rnewman:


ncalexan:
For the next person: http://developer.android.com/reference/android/content/Intent.html#ACTION_EXTERNAL_APPLICATIONS_AVAILABLE.

r=nalexander, if it needs review :)

PR #295, opened 2013-02-22
Bug 843889 - factor tabs code that is not Sync specific out of org.mozilla.gecko.sync.

ncalexan:
Let the excavations begin!

### DIFF
--- a/test/src/org/mozilla/android/sync/test/TestTabsRecord.java
+++ b/test/src/org/mozilla/android/sync/test/TestTabsRecord.java
@@ -4,6 +4,7 @@
 package org.mozilla.android.sync.test;
 
 import org.mozilla.gecko.background.db.CursorDumper;
+import org.mozilla.gecko.background.db.test.TestFennecTabsStorage;
 import org.mozilla.gecko.db.BrowserContract;
 import org.mozilla.gecko.sync.repositories.android.FennecTabsRepository;
 import org.mozilla.gecko.sync.repositories.domain.TabsRecord;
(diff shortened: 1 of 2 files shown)


rnewman:
I disagree with one particular part of this: `org.mozilla.gecko.db` is part of the Fennec tree, as is the new destination of `CursorDumper`. Firstly, our deployment scripts don't play nicely with this. Secondly, it'll be confusing as heck for other developers that _some_ files in their directories are managed upstream. Thirdly, we run the risk of collisions (like Python!).

Pick a new subpackage — org.mozilla.gecko.background works, I think — and use that as the root.

ncalexan:
Good point on the overlap with `org.mozilla.gecko.db` -- it's a pity that such a good part of the namespace is already taken :)  In fact, I think `org.mozilla.gecko.common` is too broad, and we should settle on `org.mozilla.gecko.background` for the common parts and then `org.mozilla.gecko.background.db` for our view on Fennec's datastores.  Does that sound good?

ncalexan:
By "too broad" above, I mean that we don't want to put inject `mobile/android/base/common` into the Fennec namespace -- we want to inject `mobile/android/base/background/common` (or similar).

ncalexan:
Rebased to develop and moved DB things to `org.mozilla.gecko.background.db`.

rnewman:
DO EET.

ncalexan:
Trees are closed and BMO is getting upgraded!

rnewman:
Why do you think I'm here on GitHub? :)

PR #296, opened 2013-02-23
Bug 844347 - logger

ncalexan:
Let's start building this background platform, shall we?  This separates Sync from general constants and starts building the `org.mozilla.gecko.background.common` namespace.

I'll rebase the tabs patch on top of this next week.


rnewman:
Looks good. I trust you to digest comments and take a suitable course of action!

PR #297, opened 2013-02-25
RFC? Extract BackgroundService abstract class from Announcement service classes

cpeterso:
f? @rnewman

These changes are small, but I wanted to get your feedback before submitting any bigger changes. <:) I don't have a Bugzilla bug number yet.

### DIFF
--- a/src/main/java/org/mozilla/gecko/background/BackgroundService.java
+++ b/src/main/java/org/mozilla/gecko/background/BackgroundService.java
@@ -1,0 +1,45 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.gecko.background;
+
+import org.mozilla.gecko.sync.Logger;
+
+import android.app.AlarmManager;
+import android.app.IntentService;
+import android.app.PendingIntent;
+import android.content.BroadcastReceiver;
+import android.content.Context;
+import android.content.Intent;
+import android.net.ConnectivityManager;
+import android.net.NetworkInfo;
+import android.os.Build;
+
+public abstract class BackgroundService extends IntentService {
+  private static final String LOG_TAG = BackgroundService.class.getSimpleName();
+
+  protected BackgroundService() {
+    super(LOG_TAG);
+  }
+
+  protected BackgroundService(String threadName) {
+    super(threadName);
+  }
+
+  /**
+   * Returns true if the OS will allow us to perform background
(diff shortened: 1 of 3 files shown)


ncalexan:
Looks good to me.

ncalexan:
Damn adjacent buttons.

rnewman:
Yeah, looks good to me, modulo comments above!

cpeterso:
r? @rnewman 
f? @ncalexan 
- Removed emacs comment about 4-space tabs
- Added LOG_TAG = BackgroundService.class.getSimpleName()
- Added BackgroundService.cancelAlarm()

src/main/java/org/mozilla/gecko/background/BackgroundService.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+    return networkInfo.isAvailable();
+  }
+
+  protected static final PendingIntent createPendingIntent(Context context, Class<? extends BroadcastReceiver> broadcastReceiverClass) {
+    final Intent service = new Intent(context, broadcastReceiverClass);
+    return PendingIntent.getBroadcast(context, 0, service, PendingIntent.FLAG_CANCEL_CURRENT);
+  }
+
+  protected final AlarmManager getAlarmManager() {
```
rnewman: I would suggest that this, its sibling, and perhaps `createPendingIntent` not be `final`.

rnewman:
r+ with `final` change. Thanks!

We merge like this:

```
# Assume you're on cpeterso/backgroundservice.
# Make sure you're up to date.
git rebase develop
# Make sure upstream is up to date, so that GitHub will auto-close pull reqs.
git push -f origin cpeterso/backgroundservice
# Merge.
git checkout develop
git merge --no-ff -m "Bug 845080 - Extract BackgroundService superclass. r=rnewman" cpeterso/backgroundservice
git push origin develop
```

Then land in `mozilla-inbound` and update bug accordingly.

src/main/java/org/mozilla/gecko/background/BackgroundService.java
cpeterso: @rnewman, do you mean the `getAlarmManager()` (static and non-static) methods? Some of other methods I added are also `final`. Is the Sync coding convention to not use `final` methods or classes? I noticed the Sync code uses `final` local variables.

rnewman: We use final locals to encourage/enforce immutability (which is also polite to the compiler). Final methods prevent subclasses from overriding, which in many cases hampers testing and expected reuse. 

There are places where final classes and methods are appropriate, but a nearly abstract base class probably isn't one of them :D

PR #300, opened 2013-03-01
No bug - implement preprocess.sh functionality in Python.

ncalexan:
@rnewman: take a look?  Diff reveals no changes before and after.  Small bonus: it's faster!


preprocess.ini.default
```diff
@@ -0,0 +1,36 @@
...
+# Fennec.
+
+# For standalone use.
+MOZ_UPDATE_CHANNEL = default
+MOZ_APP_DISPLAYNAME = FxSync
+MOZ_APP_VERSION = 0
+MOZ_BUILD_TIMESTAMP = %(TIMESTAMP)s
+ANDROID_CPU_ARCH = armeabi-v7a
+
+# By default, we want Android Sync to talk to a repackaged Nightly:
```
rnewman:
```
# If you're testing against a custom build, change this to, e.g.,
# ANDROID_PACKAGE_NAME = org.mozilla.fennec_username
```

preprocess.py
```diff
@@ -0,0 +1,181 @@
...
+    group.add_argument("-v", "--verbose", dest="verbose", action="store_true",
+                       default=False,
+                       help="verbose output")
+    parser.add_argument("-C", "--configs", nargs="+", metavar="INIFILE", dest="configs",
+                        default=["preprocess.ini.default", "preprocess.ini"],
+                        help="configuration file(s)")
+
+    return parser.parse_args(args)
+
+def _create_preprocess_ini(verbose):
```
rnewman, line 102: Why do this rather than copying `preprocess.ini.default`? Having duplicate content seems really bad.

indygreg:
Let me know if you need help integrating this into the m-c build system proper. That would involve things like generating the preprocess.ini during configure or something.

rnewman:
Looks good, modulo duplication. (And you probably want to put `preprocessor.in` in `.gitignore`.)

@indygreg: this runs in the standalone repo, and is necessary for standalone (non-Fennec-touching) tests to run. Early days!

preprocess.py
ncalexan, line 102: I did this because `preprocess.ini.default` is a little big and has some things that probably shouldn't be swizzled.  And by separating them, I could write targeted comments.  I was on the fence, defaulting to your perspective, and am swayed.

preprocess.py
```diff
@@ -0,0 +1,168 @@
...
+        for k in sorted(defines.keys()):
+            print("%s = %s" % (k, defines[k])) 
+
+    dot_ins = [
+        "src/main/java/org/mozilla/gecko/background/announcements/AnnouncementsConstants.java.in",
+        "src/main/java/org/mozilla/gecko/background/common/GlobalConstants.java.in",
+        "src/main/java/org/mozilla/gecko/sync/SyncConstants.java.in",
+        "src/main/java/org/mozilla/gecko/db/BrowserContract.java.in",
+        "AndroidManifest.xml.in",
+        "test/AndroidManifest.xml.in",
```
rnewman, line 135: Can we `find . -name '*.in'` instead?

rnewman:
wfm

preprocess.py
ncalexan, line 135: Yes!  I would like to pull these lists out, unify them, and feed them into new Python copy-code tooling that makes sure to never copy preprocessed files.  Soon, but not right away :)

PR #302, opened 2013-03-04
Bug 846873 - Listen for CONNECTIVITY_CHANGE events so we can cancel alarms when network is offline.

rnewman:


rnewman:
Needs tests.

PR #303, opened 2013-03-07
One step closer to a simpler setup process.

rnewman:
This should eliminate the whole 'checkout, build, mvn install' process for our unstub/stub deps.


ncalexan:
I'm pushing back on this change.  I'm on board for improving the stub/unstub deps install process, but I'd rather see us improve our tooling than moving this into source deps.  This means `mvn clean test` is cleaning and rebuilding the deps; that's lame.  Also, this pollutes the Eclipse workspace with deps that no dev should be touching.  (For example, I haven't touched these deps once in my time working on a-s.)

I'd much rather add `./tools/first_run.py` that
- checks out submodules;
- installs them in maven;
- sets up Eclipse projects;
- optionally downloads and re-packages Fennec.

I'd really rather make our other external deps (principally httpclientlib) not be source deps at all.  We build a jar in Fennec; we could build a jar in a-s as well.  (This is low priority.)

Happy to talk this out.

ncalexan:
Closed in favour of #305.

PR #304, opened 2013-03-08
Eclipsify: build Eclipse project files.

ncalexan:
When you get a moment, @rnewman.  This goes a step toward demystifying our Eclipse process and will standardize our development environments.  Documentation will need to be updated, but that will be part of the "move the wiki into the docs" project.


ncalexan:
This addresses the two review comments and changes Eclipsify into a /first run/ script that just copies `example.*` into `.*`.  Rubber stamp, @rnewman?

PR #305, opened 2013-03-13
Nalexander/mvn repo

ncalexan:
Here's a non-source version of your `deps` branch.  This kinda (not really) relies on all the other projects being updated.


rnewman:
If it works, land it!

PR #309, opened 2013-04-23
Bug 858992 - Don't sync pinned bookmarks.

rnewman:
Now with an additional test!


ncalexan:
Consider reworking some of the comments, but the test looks good and answers more questions.  Are we confident that an incoming record will not apply?  (Point me at that test.)

rnewman:
See `BookmarksTest.testForbiddenItemsAreIgnored()`, which creates a pinned item and asserts that we'll ignore it:

```
    final BookmarkRecord pinned = new BookmarkRecord("pinpinpinpin", "bookmarks", now - 1, false);
    pinned.parentID = Bookmarks.PINNED_FOLDER_GUID;
…
        assertTrue(((AndroidBrowserBookmarksRepositorySession) session).shouldIgnore(pinned));
```

`shouldIgnore` is called by `store`'s runnable, so this means that pinned items will be skipped on download.

PR #312, opened 2013-05-24
Bug 875400 - Incorporate add-ons into environment

rnewman:


ncalexan:
My head was swimming trying to keep context on the DB changes.  Can you verify that the unholy `environments` view is essentially static from here on out?

Finally, can we get a link to the document or bug describing the hash spec?  If it only matters per-client (meaning desktop or Android), that's useful to know too.  But I've heard a variety of stories around this, so good to write it down.

ncalexan:
r=me

PR #320, opened 2013-06-13
Bug 870169 - Extract reusable telemetry code from ANRReporter.java

liuche:
I realized this has zero dependency on Fennec, so it can actually live in android-sync.

Adding ANRTelemetryReporter for tracking purposes, but that needs to live in Fennec.


rnewman:
It'd be nice to have rudimentary tests for exceptions being thrown and files being written with the correct contents.

liuche:
Updated to pull #324.

PR #321, opened 2013-06-13
Bug 847178 - Avoid exceptions in CryptoRecord.fromJSONRecord.

ncalexan:
Band-aids, do you like them?!?

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/CryptoRecord.java
+++ b/src/main/java/org/mozilla/gecko/sync/CryptoRecord.java
@@ -16,6 +16,7 @@ import org.mozilla.gecko.sync.crypto.KeyBundle;
 import org.mozilla.gecko.sync.crypto.MissingCryptoInputException;
 import org.mozilla.gecko.sync.crypto.NoKeyBundleException;
 import org.mozilla.gecko.sync.repositories.domain.Record;
+import org.mozilla.gecko.sync.repositories.domain.RecordParseException;
 
 /**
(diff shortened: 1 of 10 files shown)


src/main/java/org/mozilla/gecko/sync/CryptoRecord.java
```diff
@@ -148,15 +148,29 @@ public static CryptoRecord fromJSONRecord(ExtendedJSONObject jsonRecord)
     record.guid         = id;
     record.collection   = collection;
     if (jsonRecord.containsKey(KEY_MODIFIED)) {
-      record.lastModified = jsonRecord.getTimestamp(KEY_MODIFIED);
+      Long timestamp = jsonRecord.getTimestamp(KEY_MODIFIED);
+      if (timestamp == null) {
```
rnewman, line 153: Did you check that `getTimestamp` returns null rather than throwing? I thought it threw.

rnewman:
`IOException` seems misleading to me. Would prefer

`public class RecordException extends Exception {}`.

rnewman:
Other than that, and the "make sure `getTimestamp` returns null on incorrect value" question, r+.

src/main/java/org/mozilla/gecko/sync/CryptoRecord.java
ncalexan, line 153: I did, and it manually checks with `instanceof` and returns `null`.  `getLong` casts to `Long`.

PR #322, opened 2013-06-24
Bug 886156 - Partial initialization of ProfileInformationCache.

rnewman:


rnewman:
This ensures that we don't consider ourselves restored if the on-disk data is missing `addons`. This will cause old broken files to be skipped, and a new one to be generated.

rnewman:
The situation is one in which the PIC object has Java `null` for `addons`. That'll happen when `optJSONObject` returns `null`.

As noted in previous bugs, `getJSONObject` cannot return `JSONObject.NULL`; it'll throw, and the loader method will return `false`.

ncalexan:
r=nalexander

PR #324, opened 2013-07-03
Bug 870169 - Extract reusable telemetry code from ANRReporter.java v2

liuche:
Addressed comments from pull #320 (minus perf concerns about holding onto buffers/strings/etc).

Added a few basic tests, do suggest more if needed.

### DIFF
--- a/src/main/java/org/mozilla/gecko/background/datareporting/TelemetryRecorder.java
+++ b/src/main/java/org/mozilla/gecko/background/datareporting/TelemetryRecorder.java
@@ -6,7 +6,6 @@ package org.mozilla.gecko.background.datareporting;
 
 import java.io.BufferedOutputStream;
 import java.io.File;
-import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
@@ -15,14 +14,14 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 
 import org.json.JSONObject;
+import org.mozilla.gecko.background.common.log.Logger;
 
 import android.util.Base64;
-import android.util.Log;
 
 /**
  * Writes telemetry ping to file.
  *
- * Also creates and updates a checksum for the payload to include in the ping
+ * Also creates and updates a SHA-256 checksum for the payload to include in the ping
  * file.
  *
  * A saved telemetry ping file consists of JSON in the following format,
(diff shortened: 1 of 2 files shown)


rnewman:
The reason is almost certainly to avoid having to parse that JSON when the slug is uploaded.

src/main/java/org/mozilla/gecko/background/datareporting/TelemetryRecorder.java
```diff
@@ -0,0 +1,303 @@
...
+ *
+ * A saved telemetry ping file consists of JSON in the following format,
+ *   {
+ *     "slug": "<uuid-string>",
+ *     "payload": "<escaped-json-data-string>",
+ *     "checksum": "<base64-sha-256-string>"
+ *   }
+ *
+ * This class writes first to a temporary file and then, after finishing the contents of the ping,
+ * copies that to the file specified by the caller. This is to avoid uploads of partially written
```
rnewman, line 35: Moves, not copies, right?

src/main/java/org/mozilla/gecko/background/datareporting/TelemetryRecorder.java
```diff
@@ -0,0 +1,303 @@
...
+ * ping files.
+ *
+ * The API provided by this class:
+ * startPingFile() - opens stream to a tmp File and writes the slug
```
rnewman, line 39: Writes the slug _header_, no?

src/main/java/org/mozilla/gecko/background/datareporting/TelemetryRecorder.java
```diff
@@ -0,0 +1,303 @@
...
+ * appendPayload(String payloadContent) - appends to the payload of the ping and updates the
+ *                                        checksum
+ * finishPingFile() - writes the checksum to the tmp file and copies it to the File specified by the caller.
```
rnewman, line 42: Moves.

src/main/java/org/mozilla/gecko/background/datareporting/TelemetryRecorder.java
```diff
@@ -0,0 +1,303 @@
...
+   * @throws Exception
+   *           Checked exceptions <code>NoSuchAlgorithmException</code>,
+   *           <code>UnsupportedEncodingException</code>, or
+   *           <code>IOException</code> and unchecked exception that
+   *           are rethrown to caller
+   */
+  public void startPingFile() throws Exception {
+
+    // Open stream to temporary file for writing.
+    String tmpFilename = filename + ".tmp";
```
rnewman, line 122: Nope!

http://stackoverflow.com/questions/3425906/creating-temporary-files-in-android

src/main/java/org/mozilla/gecko/background/datareporting/TelemetryRecorder.java
```diff
@@ -0,0 +1,303 @@
...
+    tmpFile = new File(parentDir, tmpFilename);
+    if (tmpFile.exists()) {
+        tmpFile.delete();
```
rnewman, line 125: Indenting.

src/main/java/org/mozilla/gecko/background/datareporting/TelemetryRecorder.java
```diff
@@ -0,0 +1,303 @@
...
+    File destFile = new File(parentDir, filename);
+    // Delete file if it exists - docs state that rename may fail if the File already exists.
+    if (destFile.exists()) {
+      destFile.delete();
+    }
+    boolean result = tmpFile.renameTo(destFile);
+    if (!result) {
+      throw new IOException("Could not move tmp file to destination.");
+    }
+    cleanUp();
```
rnewman, line 239: What does `cleanup()` do to `tmpFile`?

I suspect the answer is "it's fine, `tmpFile.exists()` returns false so it doesn't delete", but let's go ahead and null-out `tmpFile` before calling `cleanUp`, if you even care at all about discarding the checksum.

Also, shouldn't this be in a `finally` block around the whole last block?

src/main/java/org/mozilla/gecko/background/datareporting/TelemetryRecorder.java
```diff
@@ -0,0 +1,303 @@
...
+  private void cleanUp() {
+    // Discard checksum.
+    checksum.reset();
+
+    // Clean up files.
+    if (tmpFile != null && tmpFile.exists()) {
+      tmpFile.delete();
+    }
+  }
+  /**
```
rnewman, line 274: Newline before.

test/src/org/mozilla/gecko/background/datareporting/test/TestTelemetryRecorder.java
```diff
@@ -0,0 +1,166 @@
...
+
+import android.util.Base64;
+
+public class TestTelemetryRecorder extends FakeProfileTestCase {
+  private File telemetryPingDir;
+  private TelemetryRecorder telemetryRecorder;
+
+  private static final String DEST_FILENAME = "dest-filename";
+  private final String TEST_PAYLOAD1 = "{\"ver\": 1, \"measurements\":" +
+		                                 "{ \"uptime\": 24982 }, \"data\": {";
```
rnewman, line 31: Indenting.

src/main/java/org/mozilla/gecko/background/datareporting/TelemetryRecorder.java
```diff
@@ -0,0 +1,303 @@
...
+      Logger.debug(LOG_TAG, "Wrote " + footer.length + " footer bytes.");
+    } catch (UnsupportedEncodingException e) {
+      cleanUpAndRethrow("Checksum encoding exception", e);
+    } catch (IOException e) {
+      cleanUpAndRethrow("Error writing footer to stream", e);
+    } finally {
+      try {
+        outputStream.close();
+      } catch (IOException e) {
+        // Failed to close, nothing we can do.
```
rnewman, line 225: Is this a situation you just want to quietly ignore? You want to go ahead with the move?

liuche:
Closing and updating to pull #326.

src/main/java/org/mozilla/gecko/background/datareporting/TelemetryRecorder.java
liuche, line 225: I think so? The contents have been flushed, so I will null out the reference to the stream so nothing else can be written to it. There's really nothing else we can do, right?

PR #326, opened 2013-07-09
Bug 870169 - Extract reusable telemetry code from ANRReporter.java v3

liuche:
Replacing pull #324


liuche:
Addressed nits.
- Added usage of Android cachedir for files. I added a try/catch if the cacheDir fails, which defaults back to writing into the specified ping directory.
- Null out the output stream if we fail to close it, but continue with moving the file. We would have already flushed the contents of the stream to file, so we can continue with moving the file, and there's nothing we can do to close the stream if it already failed.

liuche:
I think I'd like to get this into s-c and then I'll make the changes to ANRReporter in a separate "part 2" bug.

rnewman:
r+ with that!

liuche:
Landed on m-c.

PR #327, opened 2013-07-11
Bug 888086 - Use StringBuilder in sync/Utils.java

ChrisKitching:
Finally got round to upstreaming this. Seems to pass unit testing. Hopefully I didn't maim the VCS...


rnewman:
Nits and questions:

• Not sure why there are two almost-identical commits here. The pull request should be on your branch, which should be based off upstream `develop`, so there should only be one visible. The merge commit with the review descriptor will be created when I merge the pull request (for which we don't use the GitHub UI, because it sucks).
• Indentation fixage, please.

ChrisKitching:
Drat. Attempt 2 at semi-trivial commit due shortly.

PR #329, opened 2013-07-16
Bug 893910: Reset FHR submission failure count after timeout

mcomella:
Note that this is built off of #325, which needs to be pushed before this.

@rnewman: Please review.

### DIFF
--- a/src/main/java/org/mozilla/gecko/background/healthreport/upload/SubmissionPolicy.java
+++ b/src/main/java/org/mozilla/gecko/background/healthreport/upload/SubmissionPolicy.java
@@ -72,6 +72,13 @@ public class SubmissionPolicy {
    * @return true if a request was spawned; false otherwise.
    */
   public boolean tick(final long localTime) {
+    if (localTime >= getCurrentDayResetTime()) {
+      editor()
+        .setCurrentDayResetTime(localTime + getMinimumTimeBetweenUploads())
+        .setCurrentDayFailureCount(0)
+        .commit();
+    }
+
     final long nextUpload = getNextSubmission();
 
     // If the system clock were ever set to a time in the distant future,
@@ -319,6 +326,11 @@ public class SubmissionPolicy {
     return getSharedPreferences().getInt(HealthReportConstants.PREF_CURRENT_DAY_FAILURE_COUNT, 0);
   }
 
+  // Authoritative.
+  public long getCurrentDayResetTime() {
+    return getSharedPreferences().getLong(HealthReportConstants.PREF_CURRENT_DAY_RESET_TIME, -1);
+  }
+
   /**
    * To avoid writing to disk multiple times, we encapsulate writes in a
    * helper class. Be sure to call <code>commit</code> to flush to disk!
@@ -353,6 +365,12 @@ public class SubmissionPolicy {
     // Authoritative.
     public Editor setCurrentDayFailureCount(int failureCount) {
       editor.putInt(HealthReportConstants.PREF_CURRENT_DAY_FAILURE_COUNT, failureCount);
+      return this;
+    }
+
+    // Authoritative.
+    public Editor setCurrentDayResetTime(long resetTime) {
+      editor.putLong(HealthReportConstants.PREF_CURRENT_DAY_RESET_TIME, resetTime);
       return this;
     }
 
--- a/src/main/java/org/mozilla/gecko/background/healthreport/upload/SubmissionPolicy.java
+++ b/src/main/java/org/mozilla/gecko/background/healthreport/upload/SubmissionPolicy.java
@@ -72,13 +72,6 @@ public class SubmissionPolicy {
    * @return true if a request was spawned; false otherwise.
    */
   public boolean tick(final long localTime) {
-    if (localTime >= getCurrentDayResetTime()) {
-      editor()
-        .setCurrentDayResetTime(localTime + getMinimumTimeBetweenUploads())
-        .setCurrentDayFailureCount(0)
-        .commit();
-    }
-
     final long nextUpload = getNextSubmission();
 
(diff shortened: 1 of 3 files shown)


mcomella:
#325 was merged so I rebased onto develop. I also took the liberty of making the commit messages more consistent with the code base and changed the name of the new test I added.

Should be good to review.

rnewman:
Minor nit: we put the `r=` on the merge commit when we merge, rather than on every individual commit. nbd.

src/main/java/org/mozilla/gecko/background/healthreport/upload/SubmissionPolicy.java
```diff
@@ -72,6 +72,13 @@ public SubmissionPolicy(final SharedPreferences sharedPreferences,
    * @return true if a request was spawned; false otherwise.
    */
   public boolean tick(final long localTime) {
+    if (localTime >= getCurrentDayResetTime()) {
+      editor()
+        .setCurrentDayResetTime(localTime + getMinimumTimeBetweenUploads())
```
rnewman: I think this needs a little bit of documentation, which will clarify a corner case that you might have missed.

“When a document has been uploaded (signaling the start of a new 24-hour period), or a period has otherwise started (e.g., we haven't uploaded before), we calculate and store the end of that 24-hour period.

When that time is reached, we clear our failure count.”

What you've done is imposed a hard 24-hour (actually ~25-hour, thanks to tick granularity) rotation. There's no code to set the period on success, which means the failure count loop will drift out of alignment with the upload loop, causing occasional bonus upload attempts.

I would expect to see at least two calls to `setCurrentDayResetTime`, corresponding to each of the situations in which we consider a new period to have begun: on upload success, as well as here.

src/main/java/org/mozilla/gecko/background/healthreport/upload/SubmissionPolicy.java
```diff
@@ -72,6 +72,13 @@ public SubmissionPolicy(final SharedPreferences sharedPreferences,
...
+        .setCurrentDayFailureCount(0)
+        .commit();
+    }
+
     final long nextUpload = getNextSubmission();
```
rnewman, line 75: My feeling is that the check to push out the reset time should occur as late as possible before the error count would be checked (i.e., on upload or delete). We could certainly check whether an upload is scheduled before fetching and checking that pref.

src/main/java/org/mozilla/gecko/background/healthreport/upload/SubmissionPolicy.java
rnewman: You might also consider whether some of the work can be combined with another active `editor` object — each `commit()` call writes to disk, so we want to minimize them.

mcomella:
> Ditto this. You should be able to write tests using the defaults.

Did not do this yet as I'm still waiting for a response to my reply.

@ncalexan @rnewman: Otherwise followup is ready for review.

mcomella:
Also worth noting, I reworded @rnewman 's documentation update as I didn't feel it was accurate with my conception of how the reset timer is supposed to work. Please check this over.

mcomella:
> Yes, these prefs were only configurable to keep testing doors open. We don't expect them to change and at some point we will migrate them to being actual constants.

Fixed in most recent commit.

mcomella:
Sorry, I forgot to respond to this:

> You might also consider whether some of the work can be combined with another active editor object — each commit() call writes to disk, so we want to minimize them.

We can use the same editor as `UploadDelegate` however `UploadDelegate.onSoftFailure()` tries to get the current failure count from sharedPrefs which is one of the values we're trying to commit. We can cache the reset value as an instance var in `SubmissionPolicy` and handle the cache retrieval in the accessor (i.e. `SubmissionPolicy.getCurrentDayFailureCount()`). Is this worth the added complexity?

src/main/java/org/mozilla/gecko/background/healthreport/upload/SubmissionPolicy.java
```diff
@@ -123,11 +123,23 @@ public boolean tick(final long localTime) {
...
+    // This case will occur if the nextSubmission time is not set (== -1) but firstRun is.
     if (localTime < firstRun + getMinimumTimeBeforeFirstSubmission()) {
       Logger.info(LOG_TAG, "Need to wait " + getMinimumTimeBeforeFirstSubmission() + " before first upload.");
       return false;
     }
 
+    // The first upload attempt for a given document submission begins a 24-hour period in which
+    // the upload will retry upon a soft failure. At the end of this period, the submission
+    // failure count is reset, ensuring each distinct submission attempt has a zeroed failure
+    // count. A period may also end on upload success or hard failure.
```
rnewman, line 135: This gently implies that each day tries to submit the exact same "distinct" document. That's not actually true, so I'd probably say "ensuring each day's first submission attempt has a zeroed failure count".

src/main/java/org/mozilla/gecko/background/healthreport/upload/SubmissionPolicy.java
```diff
@@ -170,6 +182,7 @@ public void onSuccess(long localTime, String id) {
         .setNextSubmission(next)
         .setLastUploadSucceeded(localTime)
         .setCurrentDayFailureCount(0)
+        .setCurrentDayResetTime(-1) // Set again on the next submission's first upload attempt.
```
rnewman: You probably want to remove, rather than set to -1; fewer magic constants. Maybe call that `clearCurrentDayResetTime()`?

rnewman:
The code looks good, but I need a little bit more brain today in order to digest the test changes.

mcomella:
Added changes for the above comments.

rnewman:
Looks good!

PR #330, opened 2013-07-17
PBKDF2 benchmarking

rnewman:
Nick, I'd appreciate the following from you:
- Observation of flaws
- Verifying whether this works for you, given that it futzes with Eclipse (you'll need to have the CDT installed)
- Variance of paths in the hard-coded commit.

I don't plan to merge this, though. (Well, we can merge the first two commits.)


ncalexan:
This has landed in various forms.

PR #332, opened 2013-07-18
Bug 870925: Add Submissions provider

mcomella:
Note that this is built off of #329.

This is a still a WIP and is being submitted for feedback purposes.


src/main/java/org/mozilla/gecko/background/healthreport/upload/AndroidSubmissionClient.java
```diff
@@ -107,20 +118,30 @@ public void upload(long localTime, String id, Collection<String> oldIds, Delegat
       long last = Math.max(getLastUploadLocalTime(), HealthReportConstants.EARLIEST_LAST_PING);
 
       if (!storage.hasEventSince(last)) {
+        // TODO: Is this actually a failure that we should count?
```
rnewman: Nope. Hard failure just means "don't bother trying again today".

src/main/java/org/mozilla/gecko/background/healthreport/upload/AndroidSubmissionClient.java
```diff
@@ -99,6 +109,7 @@ public void upload(long localTime, String id, Collection<String> oldIds, Delegat
       // out-of-process.
       HealthReportDatabaseStorage storage = EnvironmentBuilder.getStorage(client, profilePath);
       if (storage == null) {
+        // TODO: Store error in SharedPrefs so we can increment next time?
```
rnewman: File a bug.

rnewman:
f+. Lots of todos to fill in, plus rebasing on Bug 893910. Good start!

src/main/java/org/mozilla/gecko/background/healthreport/upload/AndroidSubmissionClient.java
mcomella: https://bugzilla.mozilla.org/show_bug.cgi?id=910898

src/main/java/org/mozilla/gecko/background/healthreport/upload/AndroidSubmissionClient.java
```diff
@@ -98,30 +140,49 @@ public void upload(long localTime, String id, Collection<String> oldIds, Delegat
...
-        delegate.onHardFailure(localTime, null, "Generator returned null document.", null);
+      initializeStorageForUploadProviders(storage);
+      final ProfileInformationCache profileCache = getProfileInformationCache(profilePath);
+      if (!profileCache.restoreUnlessInitialized()) {
+        Logger.warn(LOG_TAG, "Not enough profile information to compute current environment.");
         return;
       }
+      final int env = EnvironmentBuilder.registerCurrentEnvironment(storage, profileCache);
+      final int day = storage.getDay(localTime);
+      final SubmissionsStatusCounter statusCounter = new SubmissionsStatusCounter(env, day, storage);
```
rnewman: I think the scope of SSC could grow/shift a little. Consider: your uploader is doing some PIC stuff which is certainly reusable, and right here you're futzing with envs and days, using that PIC.

Indeed, the generator you just built already makes a PIC and computes the current environment, using the profile path you passed to the generator (which I think we can assume must be the same as the one we're using for recording):

```
    Environment current = EnvironmentBuilder.getCurrentEnvironment(cache);
    return generateDocument(since, lastPingTime, current);
```

To noodle on this topic a little further: perhaps a submission recorder knows about a generator, which and a generator can expose its ProfileInformationProvider and even its (optional) current environment?

And isn't that arguably a `TrackingHealthReportGenerator(storage)`?

```
/**
 * A generator that automatically records generation failures in the same storage that it used
 * to generate the document, using the current ID from its own PIC.
 * Optionally provides a wrapped RequestDelegate that will likewise record upload failures
 * if used in an upload request.
 */
public class TrackingHealthReportGenerator extends HealthReportGenerator {
  public RequestDelegate wrapRequestDelegate(RequestDelegate delegate) {
    return new UploadRequestDelegate(delegate);
  }
}
```

Simply instantiate a `TrackingHealthReportGenerator` instead of a `HRG`, and run your submissions as usual.

src/main/java/org/mozilla/gecko/background/healthreport/upload/AndroidSubmissionClient.java
```diff
@@ -98,30 +140,49 @@ public void upload(long localTime, String id, Collection<String> oldIds, Delegat
...
+
+      try {
+        incrementSubmissionAttemptCount(statusCounter);
+        final JSONObject document = generateDocument(storage, localTime, last, profilePath);
+        if (document == null) {
+          statusCounter.incrementUploadClientFailureCount();
```
rnewman: Is document generation supposed to be recorded as an upload failure?

mcomella: For this particular case of document generation, we are in the process of uploading so if this generation fails, uploading fails. Therefore, in my opinion, yes, it should be an upload failure.

src/main/java/org/mozilla/gecko/background/healthreport/upload/AndroidSubmissionClient.java
mcomella: (from IRC)

<mcomella> rnewman: I feel like the idea of the counter is necessary in other places though, like the catch statement after uploadPayload() (https://github.com/mozilla-services/android-sync/pull/332/files#diff-55c5828bd1b4c8dee19430edddb04632R181)
<mcomella> rnewman: It seems weird to me that the Generator would own the success/failure counting when it's really the uploader that should know about this
<mcomella> rnewman: Since we're counting "upload" failures, not "generation" failures
<mcomella> rnewman: I should really be getting the Environment directly in ASC.upload(), and passing that to generateDocument, rather than re-retrieving the environment, however.

What do you think?

rnewman: Then invert again: a tracker that can give you a tracking upload delegate and wrap a generator for you?

The direction I'm aiming for: the logic for tracking success and failure should be isolated to a class that implements/wraps interfaces and instruments them, rather than being spread throughout our uploader/generator code.

mcomella: Okay, I can see that. I'm not sure it's worth the time, but I'll dabble in it and find out.

mcomella:
I believe the last commit you looked at was fb3a808b58b172bfcde34bedd1e3cd820a9e30f6, "Bug 870925 - Followup: Refactor UploadRequestDelegate -> SubmissionsTracker.TrackingDelegate".

mcomella:
Sorry for the micro-commit storm. D:

For my edification, the refactor was really easy to do, but the testing refactor was pretty hairy (and a similar refactor may not be worth it in the future) - in fact, the refactor was probably significantly worse for the test's understandability.

rnewman:
That was not the most thorough review I've ever done, but lgtm.

PR #335, opened 2013-07-23
Bug 870171 - Expire old data

mcomella:
WIP - still needs the associated AndroidSubmissionClient test.

### DIFF
--- a/src/main/java/org/mozilla/gecko/background/healthreport/HealthReportDatabaseStorage.java
+++ b/src/main/java/org/mozilla/gecko/background/healthreport/HealthReportDatabaseStorage.java
@@ -1208,6 +1208,24 @@ public class HealthReportDatabaseStorage implements HealthReportStorage {
   }
 
   /**
+   * Deletes field events (rows) with dates before, but not including, the given time.
+   *
+   * @param time milliseconds since epoch. Will be converted by {@link #getDay(long)}.
+   * @return the number of field events (rows) deleted from the database.
+   */
+  public int deleteEventsBefore(long time) {
+    final SQLiteDatabase db = this.helper.getWritableDatabase();
+    final String whereClause = "date < ?";
+    final int endDay = this.getDay(time);
+    final String dayString = Integer.toString(endDay, 10);
+    final String[] whereArgs = new String[] {dayString};
+
+    int numRowsDeleted = db.delete(EVENTS_TEXTUAL, whereClause, whereArgs);
+    numRowsDeleted += db.delete(EVENTS_INTEGER, whereClause, whereArgs);
+    return numRowsDeleted;
+  }
+
+  /**
    * Retrieve a mapping from a table. Keys should be unique; only one key-value
    * pair will be returned for each key.
    */
--- a/src/main/java/org/mozilla/gecko/background/healthreport/HealthReportDatabaseStorage.java
+++ b/src/main/java/org/mozilla/gecko/background/healthreport/HealthReportDatabaseStorage.java
@@ -1220,8 +1220,35 @@ public class HealthReportDatabaseStorage implements HealthReportStorage {
     final String dayString = Integer.toString(endDay, 10);
     final String[] whereArgs = new String[] {dayString};
 
-    int numRowsDeleted = db.delete(EVENTS_TEXTUAL, whereClause, whereArgs);
-    numRowsDeleted += db.delete(EVENTS_INTEGER, whereClause, whereArgs);
+    int numRowsDeleted = 0;
+    db.beginTransaction();
+    try {
+      numRowsDeleted += db.delete(EVENTS_TEXTUAL, whereClause, whereArgs);
+      numRowsDeleted += db.delete(EVENTS_INTEGER, whereClause, whereArgs);
+      db.setTransactionSuccessful();
+    } finally {
+      db.endTransaction();
+    }
+    return numRowsDeleted;
+  }
+
+  // TODO: Name.
+  public int deleteStuffBefore(long time) {
+    final SQLiteDatabase db = this.helper.getWritableDatabase();
+    final String whereClause = "date < ?";
+    final int endDay = this.getDay(time);
+    final String dayString = Integer.toString(endDay, 10);
+    final String[] whereArgs = new String[] {dayString};
+
+    int numRowsDeleted = 0;
+    db.beginTransaction();
+    try {
+      numRowsDeleted += db.delete(EVENTS_TEXTUAL, whereClause, whereArgs);
+      numRowsDeleted += db.delete(EVENTS_INTEGER, whereClause, whereArgs);
+      db.setTransactionSuccessful();
+    } finally {
+      db.endTransaction();
+    }
     return numRowsDeleted;
   }
 
--- a/src/main/java/org/mozilla/gecko/background/healthreport/upload/AndroidSubmissionClient.java
+++ b/src/main/java/org/mozilla/gecko/background/healthreport/upload/AndroidSubmissionClient.java
@@ -111,6 +111,8 @@ public class AndroidSubmissionClient implements SubmissionClient {
         return;
       }
 
+      storage.deleteEventsBefore(localTime - HealthReportConstants.EVENT_EXISTENCE_DURATION);
+
       HealthReportGenerator generator = new HealthReportGenerator(storage);
       JSONObject document = generator.generateDocument(since, last, profilePath);
       if (document == null) {
--- a/test/src/org/mozilla/gecko/background/healthreport/test/TestHealthReportDatabaseStorage.java
+++ b/test/src/org/mozilla/gecko/background/healthreport/test/TestHealthReportDatabaseStorage.java
@@ -186,4 +186,83 @@ public class TestHealthReportDatabaseStorage extends FakeProfileTestCase {
     assertEquals(value, c.getLong(3));
     return c.moveToNext();
   }
+
+  /**
+   * Returns a storage instance prepopulated with dummy data to be used for testing.
+   *
+   * Note: Editing this data directly will cause tests relying on it to fail. To add additional
+   * data, two possiblilites are 1) this method is wrapped and the data is added to the returned
+   * object, or 2) this method takes an "version" argument with new data additions running only if
+   * the version is greater than some value. Once this is implemented, this comment can be
+   * removed.
+   *
+   * XXX: This is used in lieu of subclassing TestHealthReportDatabaseStorage in an inner class
+   * and prepopulating the storage instance in setUp() because the test runner was unable to find
+   * the inner class in testing.
+   */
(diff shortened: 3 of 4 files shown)


src/main/java/org/mozilla/gecko/background/healthreport/HealthReportDatabaseStorage.java
```diff
@@ -1208,6 +1208,24 @@ public Cursor getEventsSince(long time) {
...
+   */
+  public int deleteEventsBefore(long time) {
+    final SQLiteDatabase db = this.helper.getWritableDatabase();
+    final String whereClause = "date < ?";
+    final int endDay = this.getDay(time);
+    final String dayString = Integer.toString(endDay, 10);
+    final String[] whereArgs = new String[] {dayString};
+
+    int numRowsDeleted = db.delete(EVENTS_TEXTUAL, whereClause, whereArgs);
+    numRowsDeleted += db.delete(EVENTS_INTEGER, whereClause, whereArgs);
```
rnewman: Might be good to put these two lines in a transaction. No sense making an expensive operation even more expensive...

test/src/org/mozilla/gecko/background/healthreport/test/TestHealthReportDatabaseStorage.java
```diff
@@ -187,4 +186,105 @@ private static boolean assertRowEquals(Cursor c, int day, int env, int field, in
     assertEquals(value, c.getLong(3));
     return c.moveToNext();
   }
+
+  /**
+   * Returns a storage instance prepopulated with dummy data to be used for testing.
+   *
+   * Note: Editing this data directly will cause tests relying on it to fail. To add additional
+   * data, two possiblilites are 1) this method is wrapped and the data is added to the returned
```
rnewman: Spellcheck!

test/src/org/mozilla/gecko/background/healthreport/test/TestHealthReportDatabaseStorage.java
```diff
@@ -187,4 +186,105 @@ private static boolean assertRowEquals(Cursor c, int day, int env, int field, in
...
+   * object, or 2) this method takes an "version" argument with new data additions running only if
+   * the version is greater than some value. Once this is implemented, this comment can be
+   * removed.
+   *
+   * XXX: This is used in lieu of subclassing TestHealthReportDatabaseStorage in an inner class
+   * and prepopulating the storage instance in setUp() because the test runner was unable to find
+   * the inner class in testing.
```
rnewman, line 201: Was your inner class `static`?

http://stackoverflow.com/questions/8758294/test-cases-in-inner-classes-with-junit

test/src/org/mozilla/gecko/background/healthreport/test/TestHealthReportDatabaseStorage.java
```diff
@@ -187,4 +186,105 @@ private static boolean assertRowEquals(Cursor c, int day, int env, int field, in
...
+        fields.add(new FieldSpec("counterField", Field.TYPE_INTEGER_COUNTER));
+        fields.add(new FieldSpec("discreteField", Field.TYPE_STRING_DISCRETE));
+        fields.add(new FieldSpec("lastField", Field.TYPE_STRING_LAST));
+        return fields;
+      }
+    }, new MeasurementFields() {
+      @Override
+      public Iterable<FieldSpec> getFields() {
+        ArrayList<FieldSpec> fields = new ArrayList<FieldSpec>();
+        fields.add(new FieldSpec("counterField", Field.TYPE_INTEGER_COUNTER));
```
rnewman: Perhaps give these two sets of fieldspecs different names?

`b_discrete_string`?

src/main/java/org/mozilla/gecko/background/healthreport/upload/AndroidSubmissionClient.java
```diff
@@ -111,6 +111,8 @@ public void upload(long localTime, String id, Collection<String> oldIds, Delegat
         return;
       }
 
+      storage.deleteEventsBefore(localTime - HealthReportConstants.EVENT_EXISTENCE_DURATION);
```
rnewman, line 114: I'm in two minds about this.

Firstly, it should be approximately equivalent to prune immediately after submission, but safer: no chance of an OOM during deletion, etc.

Secondly, do we want to bomb out if this deletion fails for some reason, or should this be wrapped in a `try` block?

src/main/java/org/mozilla/gecko/background/healthreport/HealthReportDatabaseStorage.java
```diff
@@ -1208,6 +1208,24 @@ public Cursor getEventsSince(long time) {
   }
 
   /**
+   * Deletes field events (rows) with dates before, but not including, the given time.
```
rnewman, line 1211: There's something you missed here, and that's environment cleanup.

I suggest the following logic:
- If there are any rows to delete
- Find their environments (`SELECT DISTINCT env FROM events... WHERE date < ?`)
- Delete the rows
- Delete any environments that are both in the set and are no longer used by any events.

I suggest this complicated version to avoid deleting the current environment. An equivalent is this:

```
DELETE FROM environments
WHERE (SELECT COUNT(*) FROM events_view WHERE date >= ? 
       AND events_view.id = environments.id) = 0
AND id IN (SELECT DISTINCT env FROM events_view WHERE date < ?);

DELETE FROM events_integer WHERE date < ?;
```

(though you might want to replace the view reference with a `UNION ALL` to leverage indices.)

This will only work if I remembered to put `ON DELETE CASCADE` in the environments table definition. If I didn't, you'll need to do this in parts.

Extra points if you figure out a way to delete old environments that were never used...

test/src/org/mozilla/gecko/background/healthreport/test/TestHealthReportDatabaseStorage.java
mcomella, line 201: Yeah, it was.

Eclipse would notice the class if I _only_ ran the tests for the TestHealthReportDatabaseStorage file (it would prompt me to run the tests for either the inner or outer class) but when running them in aggregation, neither eclipse nor maven ran the tests from the inner class.

src/main/java/org/mozilla/gecko/background/healthreport/upload/AndroidSubmissionClient.java
mcomella, line 114: > Firstly, it should be approximately equivalent to prune immediately after submission, but safer: no chance of an OOM during deletion, etc.

My intent with placing it before was for performance but I realized there is an index on the date field for the event tables, which should make the events query (>= date) quick not matter where the pruning took place. Would you agree (my SQL skills are a bit rusty)?

However, I'm not sure I understand the OOM case - we still need to maintain the same references when we delete before as with after (namely a `HealthReportDatabaseStorage`instance and a `ContentProviderClient` instance to get the storage). Can you explain why this could OOM? Is it because an OOM exception would likely be thrown on the document generation query if it was first?

Otherwise, I agree the placement should not matter.

> Secondly, do we want to bomb out if this deletion fails for some reason, or should this be wrapped in a try block?

I've discovered that since we run in the same process as Fennec, if we bomb out, we'll kill Fennec too ([reference](https://bugzilla.mozilla.org/show_bug.cgi?id=740731#c13)).

Thus, we should try/catch in any case, but it's probably good to log this error in some way because (and correct me if I'm wrong) a delete failing is probably either database corruption or programmer error, the latter which we can fix.

rnewman, line 114: Re index: depending on the frequency of deletion, even if there wasn't an index, we're talking about skipping approx one (or five, or seven) record(s) in 180, so it doesn't matter too much where the pruning happens!

My concern about OOM etc.: `deleteEventsBefore` is going to allocate objects, and it has other ways of failing (some memory-related, such as paging in indices, and some just Java).

If our pruning fails for any reason, we'll never submit a document. (And even with a `try..catch`, we can end up in a situation where the upload fails; segfaults, for example.)

So this is a case of prioritization: if submission succeeds today but pruning fails, great; pruning will probably succeed tomorrow. If pruning fails and we don't submit, then sadface.

Yes, we should log, and eventually dump into the FHR error collector itself.

src/main/java/org/mozilla/gecko/background/healthreport/HealthReportDatabaseStorage.java
mcomella, line 1211: > (though you might want to replace the view reference with a `UNION ALL` to leverage indices.)

To make sure I understand, you mean to perform this `SELECT` statement on each events table separately and `UNION ALL` the results?

> This will only work if I remembered to put `ON DELETE CASCADE` in the environments table definition. If I didn't, you'll need to do this in parts.

The event tables' foreign key reference to the environments table have it, which contradicts what you said but I think that's the way it's supposed to be for this to work.

> Extra points if you figure out a way to delete old environments that were never used...

I came up with:

```
DELETE FROM environments
WHERE id IN
        (SELECT id
        FROM environments LEFT OUTER JOIN events_view
        ON environments.id = events_view.env
        WHERE date IS NULL AND id != ?);
```

Where `?` refers to the current environment's ID. Note that this is untested (and potentially syntactically incorrect). I tried pretty hard (a couple of hours) to combine your query and mine in an efficient way, but to no avail. If an `OR` is appended to your query and everything up to `id` removed from mine, they can be combined in a rather trivial, silly way though.

Presumably the left outer join will be significantly more expensive than the joins in your query because the indices cannot be used. As such, this orphaned environment query may not be something we want to run every time we delete from the database.

It might be possible to mitigate this expensiveness by running each `SELECT` individually on a temporary table with the results of the `LEFT OUTER JOIN` so that the DB could potentially do all the operations in one pass (assuming this is possible - I'm not very proficient in my DB knowledge).

Alternatively, we can keep some kind of "unused env" metadata table, but that will increase code/schema complexity and make the recording of measurements more expensive (but potentially less impactful than the expensive join above).

Without benchmarking, I personally prefer the idea of deleting the unused environments less often than we delete old events (or do both at the same time but fairly infrequently) as most environments (at least with the current environment representation) will have measurements and thus not be unused. Would you agree?

rnewman, line 1211: I'm 99% sure that the right index usage will occur if you use the view. But be prepared for it to be inexplicably slow! `EXPLAIN QUERY PLAN` is your friend.

For deleting empty environments:

```
  DELETE FROM environments WHERE
  id != ? AND
  id NOT IN (
    SELECT DISTINCT env FROM events_view
  )
```

should do the trick no? "Delete environments that aren't in the set of used environments".

This should be fairly cheap: it's an easy index walk and then an easy table scan.

mcomella:
Closing in favor of #344, which is a large rewrite over this. This is done instead of rewriting history to preserve the comments above.

PR #336, opened 2013-07-31
Bug 899840 - Use sort for consistent services makefile ordering.

mcomella:


ncalexan:
r=nalexander.  Good comment about locale, and good work finding a zero-churn solution.  No need to include the bug number -- that's what git blame is for.  (Our views on this have changed over time, so there are some places we do that in the tree.  Mea culpa.)  I'm going to merge this to develop.

Thanks for squashing issues you find developing on Linux!  me, rnewman, and liuche develop on Macs so we haven't been as vigilant as we could have been.

ncalexan:
https://github.com/mozilla-services/android-sync/commit/9ed44571a0b76a8b3b8185bdd4ece73d8144010a

PR #342, opened 2013-08-02
Bug 900289 - Add foreign key pragma to FHR database and associated tests

mcomella:


mcomella:
Additional concern: What assumptions are we making about null values? HealthReportDatabaseStorage.ensureMeasurementInitialized never checks for a null string which I imagine would put null into the database. This test (and subsequent fix) can be added to this series of commits.

mcomella:
Additional concern 2: Do we want to verify that old databases don't violate constraints? We can dbVersion++ and remove any nulls from violations there.

mcomella:
Additional concern 3: Performance implications of foreign keys. I imagine it's negligible, but it might be worth investigating.

rnewman:
You could skip the `PRAGMA` for read-only DB connections. That would help with any perf issue. Write load is low-frequency enough that it's not a concern to me.

rnewman:
And yeah, we should probably discard any inconsistent data in a migration.

src/main/java/org/mozilla/gecko/background/healthreport/HealthReportDatabaseStorage.java
```diff
@@ -347,6 +348,13 @@ public void onCreate(SQLiteDatabase db) {
       }
     }
 
+    @Override
+    public void onOpen(SQLiteDatabase db) {
+      if (!db.isReadOnly()) {
```
rnewman, line 358: Heh, I guess you already did what I suggested :P

src/main/java/org/mozilla/gecko/background/healthreport/HealthReportDatabaseStorage.java
```diff
@@ -1031,7 +1039,13 @@ private void recordDailyLast(int env, int day, int field, Object value, String t
       v.put("env", env);
       v.put("field", field);
       v.put("date", day);
-      db.insert(table, null, v);
+      try {
+        db.insertOrThrow(table, null, v);
+      } catch (SQLiteConstraintException e) {
+        throw new IllegalStateException("Event did not reference existing an environment or field.", e);
+      } catch (Exception e) {
+        Logger.error(LOG_TAG, "Unknown exception thrown while recording daily last value.", e);
```
rnewman: I'd rather have the original exception propagate rather than be logged; just drop this catch clause.

src/main/java/org/mozilla/gecko/background/healthreport/HealthReportDatabaseStorage.java
```diff
@@ -1063,7 +1077,13 @@ private void recordDailyDiscrete(int env, int day, int field, Object value, Stri
...
     final SQLiteDatabase db = this.helper.getWritableDatabase();
     putValue(v, value);
-    db.insert(table, null, v);
+    try {
+      db.insertOrThrow(table, null, v);
+    } catch (SQLiteConstraintException e) {
+      throw new IllegalStateException("Event did not reference existing an environment or field.", e);
+    } catch (Exception e) {
+      Logger.error(LOG_TAG, "Unknown exception thrown while recording daily discrete value.", e);
+    }
```
rnewman, line 1105: Same.

src/main/java/org/mozilla/gecko/background/healthreport/HealthReportDatabaseStorage.java
```diff
@@ -1133,7 +1153,13 @@ public void incrementDailyCount(int env, int day, int field, int by) {
...
       v.put("field", field);
       v.put("date", day);
-      db.insert(EVENTS_INTEGER, null, v);
+      try {
+        db.insertOrThrow(EVENTS_INTEGER, null, v);
+      } catch (SQLiteConstraintException e) {
+        throw new IllegalStateException("Event did not reference existing an environment or field.", e);
+      } catch (Exception e) {
+        Logger.error(LOG_TAG, "Unknown exception thrown while incrementing daily count.", e);
+      }
```
rnewman, line 1179: Same.

test/src/org/mozilla/gecko/background/healthreport/test/MockHealthReportDatabaseStorage.java
```diff
@@ -39,4 +50,140 @@ public SQLiteDatabase getDB() {
...
+  /**
+   * A storage instance prepopulated with dummy data to be used for testing.
+   *
+   * XXX: This is used in lieu of subclassing TestHealthReportDatabaseStorage in an inner class
+   * and prepopulating the storage instance in setUp() because the test runner was unable to find
+   * the inner class in testing.
+   *
+   * TODO: Modifying this data directly will cause tests relying on it to fail so a versioned
+   * constructor should be added where additional (or entirely different) data is used for each
+   * version.
```
rnewman: Can we address the XXX/TODO somehow before landing?

test/src/org/mozilla/gecko/background/healthreport/test/TestHealthReportDatabaseStorage.java
```diff
@@ -187,4 +198,171 @@ private static boolean assertRowEquals(Cursor c, int day, int env, int field, in
...
+   */
+  public void testInsertions() throws Exception {
+    final PrepopulatedMockHealthReportDatabaseStorage storage =
+        new PrepopulatedMockHealthReportDatabaseStorage(context, fakeProfileDirectory);
+    assertNotNull(storage);
+  }
+
+  private int getNonExistentID(SQLiteDatabase db, String table) {
+    final String[] columns = {"id"};
+    final Cursor c = db.query(table, columns, null, null, null, null, "id DESC");
```
rnewman: I know this is test code, but `LIMIT 1` is your friend.

Or...

```
SELECT MAX(id) + 1 FROM $table
```

test/src/org/mozilla/gecko/background/healthreport/test/TestHealthReportDatabaseStorage.java
```diff
@@ -187,4 +198,171 @@ private static boolean assertRowEquals(Cursor c, int day, int env, int field, in
...
+    final int discreteFieldID = storage.getField(storage.measurementNames[0], storage.measurementVers[0],
+        storage.fieldSpecContainers[0].discrete.name).getID();
+
+    final int nonExistentEnvID = getNonExistentID(db, "environments");
+    final int nonExistentFieldID = getNonExistentID(db, "fields");
+
+    try {
+      storage.incrementDailyCount(nonExistentEnvID, storage.getToday(), counterFieldID);
+      fail("Should throw - event_integer(env) references environments(id), which is given as a non-existent value.");
+    } catch (IllegalStateException e) { }
```
rnewman, line 324: Note that the change I noted that muffles exceptions will cause this test to pass if anything other than a `SQLConstraintException` is thrown, which is a footgun...

rnewman:
This looks good to me. Take a look at migration/cleanup next, 'cos we'll want to land those together.

test/src/org/mozilla/gecko/background/healthreport/test/MockHealthReportDatabaseStorage.java
mcomella: I can add the versioning mechanic (the TODO), but there isn't really anything I can do about the XXX. Should I just drop the comment? I think it's less relevant now that this is more than a method.

mcomella: Dropped the comment and added the versioning mechanic in 276944c2c8d98ccd741da72a35b80e30b7e6efc5, but there unfortunately isn't any data that uses it yet.

test/src/org/mozilla/gecko/background/healthreport/test/TestHealthReportDatabaseStorage.java
mcomella, line 324: Just to make sure I understand, you're saying if I didn't remove the

```
} catch (Exception e) { ... }
```

from the above code, this test would pass when other errors are thrown? Meaning this comment is for my edification (and thank you for that!) and there are no changes to be made?

rnewman, line 324: Correct.

test/src/org/mozilla/gecko/background/healthreport/test/MockHealthReportDatabaseStorage.java
rnewman: Yeah, just drop the comment or rewrite it to be non-XXXey :D

mcomella:
Also, before I forget, please comment on https://github.com/mozilla-services/android-sync/pull/342#issuecomment-22042267. I feel like assertions would be useful here but I don't believe they're enabled by our build system. :(

rnewman:
I'm not sure what you mean by assertions in this context; what can we check at build time?

mcomella:
Like

```
assert measurementNames != null;
```

But I do not believe assertions are enabled in our build system. Alternatively,

```
if (measurementNames == null) {
    throw new IllegalStateException("measurementNames should not be null.");
}
```

Or better yet,

```
public void ensureMeasurementsInitialized(String[] measurementNames, ...) {
    ensureNonNull(measurementNames);
    // ...
}

private void ensureNonNull(Object obj) {
    if (obj == null) {
        throw new IllegalStateException("obj was null!");
    }
}
```

It'd probably be good to check for these things rather than assuming they'll never be null, though at the cost of some code complexity so I wanted to know if such checks are worthwhile.

rnewman:
Yeah, the typical approach is to use exceptions during validation. E.g., https://github.com/mozilla-services/android-sync/blob/develop/src/main/java/org/mozilla/gecko/sync/Utils.java#L209

If doing something will break things, the only three options are:
- Throw (validate)
- Ignore/log (leads to problems later, and relies on a web of contracts: what happens when I later try to retrieve my measurement by name?)
- Specify preconditions — e.g., in Javadoc,

```
@param measurementNames an array of non-null <code>String</code> instances.
@throw IllegalArgumentException if a <code>null</code> measurement name is provided.
```

In richer languages we wouldn't have this problem.

mcomella:
In response to argument validation, without being overzealous, nothing I touched here seems to be a prime candidate for such validation but I'll keep an eye open in the future for places where it's applicable. Thanks for the edification.

mcomella:
@rnewman: Ready for review if you didn't get the BZ notification.

mcomella:
@rnewman: Ready for followup review.

rnewman:
LGTM.

PR #344, opened 2013-08-06
Bug 870171 - Expire old data (2)

mcomella:
Note that this is built on #342 which must be pushed first.

Replaces #335 (for more history, see old comments there).


mcomella:
Concerns:
- The tests are redundant - the larger methods are largely copy-paste of the smaller methods they call. Is there any good way to alleviate this? This is especially pertinent because this test class now takes anywhere from 10 to 30 seconds to run.
- Was I overzealous in testing? I felt like this took me much longer than it probably should have.

TODO:
- Benchmarking and `EXPLAIN` query to ensure index use, as per https://github.com/mozilla-services/android-sync/pull/335#discussion_r5490512.
- Use deleteDataBefore in the code (and add associated tests).
- Periodic pruning and vacuuming, as per https://bugzilla.mozilla.org/show_bug.cgi?id=870171#c4 (but in a separate bug?)

rnewman:
lgtm.

mcomella:
Rebased - previous LGTM up to a898254086a65a6f737a43fce0a205548085cdb2 (Part 5 review comment).

ncalexan:
Parts 6 through 12 look okay to me.  I would strip some of the "deep Android" testing; it's not all that high value, and it's hard to get Android testing to play nicely.  (Witness the intermittent problems with the Account tests.)

To address your question about over-zealous testing: yes; feel free to push the meat and follow-up with tests after preliminary review.  This really needs to be hand-tested with QA's help, so design for easy toggle-ing/wait period-ing rather than automated testing in these cases.

mcomella:
> This really needs to be hand-tested with QA's help, so design for easy toggle-ing/wait period-ing rather than automated testing in these cases.

Can you provide an example or elaborate more on "toggle-ing/wait period-ing"? Are you saying the Upload and Prune services should be able to be triggered at will so QA can hand test the before and after state?

rnewman:
The one thing I'd add is: we have a hook for memory usage.

```
src/main/java/org/mozilla/gecko/background/healthreport/HealthReportProvider.java
101:  public void onLowMemory() {
```

It would be worth adding a comment there to explain why we don't prune the DB. Someone will eventually put two and two together and get three.

mcomella:
> The one thing I'd add is: we have a hook for memory usage.
> It would be worth adding a comment there to explain why we don't prune the DB.

Potentially dumb question but why would someone think to prune the DB on low memory?

rnewman:
Re pruning on low memory: 'cos DB connections use memory, there's not a great distinction drawn between device storage and RAM, page caching, yadda yadda.

And when I came to review, it occurred to me "huh, maybe we should be pruning on or after a memory pressure warning", and I had to think about it to realize it would (probably) be a bad idea. If I thought about it, someone else will!

mcomella:
Sure, 2a31f5da5ccc747b76b1134ccdd314a6e227052c.

The only connection I could really understand is the ambiguity between RAM and disk space so that's what I wrote about. Tell me if you want any additional information added to the comment.

mcomella:
@ncalexan: ping on my replies to the review comments for Parts 6-11.

ncalexan:
In response to "This really needs to be hand-tested with QA's help, so design for easy toggle-ing/wait period-ing rather than automated testing in these cases.":

The prune service will run periodically.  Better to make the period a preference that can be easily poked at (using the Background Services Lab, for example) than extensively testing all sorts of automated business.  That way, a human can just be sure the events fire, alarms go off, etc when necessary.  It's a judgement call about which to do when.

ncalexan:
re: testing: lgtm.

mcomella:
Self-note TODO:
- ~~Get response for [last testing comment](https://github.com/mcomella/android-sync/commit/a3e52408c1fe23161aea46fd92a18fdf5f59aecd#commitcomment-4036585)~~
- Figure out how DB upgrades can fail (see [rnewman's comment](https://github.com/mcomella/android-sync/commit/80c332191dee906bb37df8c52205a7c7e89eb5d2#commitcomment-4007393))
- Test vacuuming on low-end devices to ensure we don't OOM or anything terrible (wesj has a Galaxy Chat)
- ~~Look into [DB efficiency comment](https://github.com/mcomella/android-sync/commit/81144019efdf911d65bb44e6ead7084bc1853c0e#src-main-java-org-mozilla-gecko-background-healthreport-healthreportdatabasestorage-java-P13)~~
- Benchmark to find decent values for HealthReportConstants
- ~~Finish `attemptPruneBySize`, `attemptVacuum`, and `attemptExpiration` tests.~~
- ~~Check for TODOs - I think I left some in there~~
- ~~`PrunePolicyDatabaseStorage` tests.~~

mcomella:
I rebased on 23-26 because I'm a jerk :astonished: (i.e. I moved `PrunePolicy.getCurrentEnvironmentID()` to `AndroidPrunePolicyStorage` because it made the design a bit cleaner). Sorry if you were already reviewing but just make sure nothing too terrible changed! :)

rnewman:
Ship it ship it

rnewman:
Code looks good; let's just get a good handle on the constants.

mcomella:
(Rebased and deleted vacuuming based on fragmentation count and associated page size changes).

rnewman:
Ship it

mcomella:
Rebased. There were a fair number of conflicts with the `GlobalConstants` changes so each revision may not pass the tests or compile correctly, though everything works correctly at HEAD.

mcomella:
Collapsed followups - it's unfortunately too hard to put them back into the original commits.

mcomella:
shipped

PR #346, opened 2013-08-06
No Bug - Add HTML report to the integration testing information.

mcomella:
I find this useful when I run via the command line, rather than having to sort through the shell scrollback, so perhaps someone else will too.

@ncalexan: Please review.


ncalexan:
Looks good!

PR #350, opened 2013-09-03
Bug 912275 - scrypt helper client.

ncalexan:
Easing back into this whole "work" thing.


rnewman:
Looks good to me.

Perhaps we need a "Live" test project?

ncalexan:
We're not intending to use this on Android at all.

PR #351, opened 2013-09-04
Bug CRYPTO - Add Scrypt and PBKDF2 interfaces.

ncalexan:
This also adds testing only implementations backed by SpongyCastle.

@rnewman, this is an alternate path toward landing some sort of native crypto that will be performant on actual devices.  This unblocks me so that I can implement against the auth server, while allowing a reasonable upgrade path as we experiment with better implementations.  What say you?

### DIFF
--- a/android-sync-app/pom.xml
+++ b/android-sync-app/pom.xml
@@ -72,6 +72,12 @@
       <version>4.1.21</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.madgag</groupId>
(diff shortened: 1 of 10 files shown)


rnewman:
Did you see

https://github.com/mozilla-services/android-sync/blob/335141b8b32ba8b817b3fbde9d0bf3e55f93bea5/src/main/java/org/mozilla/gecko/sync/crypto/PBKDF2.java

? (Part of #330)

ncalexan:
This has all landed in various forms.

PR #352, opened 2013-09-09
First version of FxAccountClient.

ncalexan:
@rnewman: this isn't quite done, but it's ready for feedback.  I intend to give it a documentation pass, and to factor out, test, and make more strict most of the parsing code.  I don't want to work on this anymore right now and am cycling onto the next ticket :)

Oops, I forgot: I think the FxAccount/SRPSession can be simplified and possibly combined, but I want to write some of the other pieces to see what feels right.


rnewman:
The first half of this is now covered by my native crypto pull req. Will finish cleaning up later.

ncalexan:
Parts of this have landed; parts have been changed by protocol; and parts won't land.

PR #353, opened 2013-09-12
Bug 915312 — Native PBKDF2 and SCrypt; Bug 905433 — Key stretching; WIP.

rnewman:
Nick: your SRP patches are at the end of this sequence. This pull request is for the numbered bugs at the start.

Note that this goes straight for the jugular, using native code for both PBKDF2 and SCrypt. All of the active tests pass, though the key stretching bug has a lot of commented-out tests.

Follow-ups are needed for the following:

• Somehow generating the .cproject file to be user-specific, or otherwise solving that issue.
• Ensuring that we're happy with the state of licensing.
• Landing the native code in m-c.

Note also that I killed some of the spurious dependencies that had crept in: Spongy Castle, of course, but also Jackson and the v4 compat library. I trust that these are present in other pull requests if needed — tests pass without them.

Finally: there's a separate branch (`rnewman/srp-native-openssl`) that adds OpenSSL to this mix. It buys a 15% or more speedup on PBKDF2, but I haven't spent the time to pull out the PBKDF2 code yet. This'll do for now.


ncalexan:
This has all landed in various forms.

PR #355, opened 2013-09-23
No bug - Add info that maven v3.1.0 is broken.

mcomella:
Perhaps this info could be in a better place but I wouldn't be sure where.


ncalexan:
lgtm.

PR #358, opened 2013-09-25
Bug 799732 - token server client

ncalexan:
This builds on 799734 so that we can test it against live servers.  (There are some static tests for good measure, too.)


rnewman:
This looks good to me (albeit not the deepest review I've ever done!)

PR #359, opened 2013-09-25
Bug 920674 - Add log statements.

mcomella:


mcomella:
Enough log statements? It should be - the log statements added with the initial push cover the rest of it - it was just difficult to tell what was happening when prune-by-size was called and neither events or envs were pruned, or when cleanup actually occurred.

mcomella:
On second thought, cleaning up a few others that were a bother.

mcomella:
@ncalexan: ^ Just making sure you see this.

mcomella:
r+ via irc:

> nalexander> mcomella: yes, I saw your comment about logging.  roll on.

PR #361, opened 2013-09-27
Bug 892025 - Implement FxAccountClient.

ncalexan:
This is substantively ready.  There's some commenting to be done, to explain the unusual delegate structure, but we're mostly there.  Ready for first review, @rnewman.


ncalexan:
Bug 889749 is just for testing; I'll post an elm-ready version ASAP.

rnewman:
This looks pretty good; see comments.

I would like a second pair of eyes to evaluate the actual protocol-related bytefoolery in the meaty commit. That can be a real sec review, or it can be warner reading it and making sure it does the right thing. Your choice.

ncalexan:
This landed a while back.

PR #364, opened 2013-10-10
Bug 902560 - FakeProfileTestCase deletes testing directory if it exists

mcomella:
...and properly cleans up after itself.

@ncalexan: r? (will needinfo on bz)


mcomella:
Rebased, which removed r+. Naming nit fixed in cd708feb2768f5b246eae5e2ca2f396c2625b4f1.

Also added 92e2871c538968095f503ff7ba0d1667dd6228b7 which is a correction and should be uncontroversial.

PR #367, opened 2013-10-21
Bug 929066 - Allow for local clock skew.

ncalexan:
This is a stop-gap, as described in the ticket. It merely issues certs 20 seconds in the past, and assertions 10 seconds in the past. (Assertions must be issued after certificates.) This works around the little bit of local clock skew that's been frustrating me.

@rnewman, trivial review?


ncalexan:
Closing since we're doing this right(er) now.

PR #368, opened 2013-10-21
Bug 889749 - ciphersuite.

ncalexan:
Here's how I'd like to handle this short-term: I want to land this on elm _only_, 'cuz I don't want to pay the testing cost for doing this for Sync just yet.  It's working well for me in practice and will let us move ahead with engineering milestone 0.

The attached (live, sigh) test will fail against develop, but pass with this patch.

@rnewman: r?


rnewman:
Do eet.

PR #371, opened 2013-10-31
Bug 933445 - Incorrect paths for android-services.mozbuild with fennec-code-copy.sh on Linux

mcomella:


mcomella:
Verified by nalexander on IRC.

PR #372, opened 2013-11-15
Bug 918012 - sync with token server

ncalexan:
@rnewman: The first 7 parts are ready for review.  These patches split out pieces that in turn make it easy to build a new `FxAccountSyncAdapter` similar to `SyncAdapter` but intended for Firefox Accounts auth.  That work just needs cleaning before it goes up for review.

### DIFF
--- a/src/main/java/org/mozilla/gecko/fxa/authenticator/FxAccountAuthenticator.java
+++ b/src/main/java/org/mozilla/gecko/fxa/authenticator/FxAccountAuthenticator.java
@@ -7,6 +7,7 @@ package org.mozilla.gecko.fxa.authenticator;
 import org.mozilla.gecko.AppConstants;
 import org.mozilla.gecko.background.common.log.Logger;
 import org.mozilla.gecko.fxa.FxAccountConstants;
+import org.mozilla.gecko.fxa.activities.FxAccountSetupActivity;
 import org.mozilla.gecko.fxa.sync.FxAccount;
 import org.mozilla.gecko.sync.Utils;
 
@@ -18,7 +19,6 @@ import android.accounts.NetworkErrorException;
 import android.content.ContentResolver;
 import android.content.Context;
 import android.content.Intent;
-import android.net.Uri;
 import android.os.Bundle;
 
 public class FxAccountAuthenticator extends AbstractAccountAuthenticator {
@@ -86,11 +86,7 @@ public class FxAccountAuthenticator extends AbstractAccountAuthenticator {
       return res;
     }
 
-    // Start Fennec at about:accounts page.
-    Intent intent = new Intent(Intent.ACTION_VIEW);
-    intent.setClassName(AppConstants.ANDROID_PACKAGE_NAME,
-        AppConstants.ANDROID_PACKAGE_NAME + ".App");
-    intent.setData(Uri.parse("about:accounts"));
+    Intent intent = new Intent(context, FxAccountSetupActivity.class);
     res.putParcelable(AccountManager.KEY_INTENT, intent);
     return res;
   }
--- a/src/main/java/org/mozilla/gecko/fxa/authenticator/FxAccountAuthenticator.java
+++ b/src/main/java/org/mozilla/gecko/fxa/authenticator/FxAccountAuthenticator.java
@@ -15,10 +15,19 @@ import android.accounts.AccountManager;
 import android.accounts.NetworkErrorException;
 import android.content.ContentResolver;
 import android.content.Context;
+import android.content.Intent;
+import android.net.Uri;
 import android.os.Bundle;
 
 public class FxAccountAuthenticator extends AbstractAccountAuthenticator {
   public static final String LOG_TAG = FxAccountAuthenticator.class.getSimpleName();
+
+  public static final String JSON_KEY_UID = "uid";
+  public static final String JSON_KEY_SESSION_TOKEN = "session_token";
+  public static final String JSON_KEY_KA = "kA";
+  public static final String JSON_KEY_KB = "kB";
+  public static final String JSON_KEY_IDP_ENDPOINT = "idp_endpoint";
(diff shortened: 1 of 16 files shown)


rnewman:
Parts 1-7 can hit m-c.

ncalexan:
The previously reviewed parts have already hit m-c.  Here are the last bits to Sync against an Fx Account.

src/main/java/org/mozilla/gecko/fxa/authenticator/FxAccountAuthenticator.java
```diff
@@ -44,31 +86,12 @@ public Bundle addAccount(AccountAuthenticatorResponse response,
...
-      ContentResolver.setIsSyncable(account, authority, 1);
-    }
-
-    res.putString(AccountManager.KEY_ACCOUNT_NAME, account.name);
-    res.putString(AccountManager.KEY_ACCOUNT_TYPE, account.type);
+    // Start Fennec at about:accounts page.
+    Intent intent = new Intent(Intent.ACTION_VIEW);
+    intent.setClassName(AppConstants.ANDROID_PACKAGE_NAME,
+        AppConstants.ANDROID_PACKAGE_NAME + ".App");
+    intent.setData(Uri.parse("about:accounts"));
```
rnewman: Perhaps worth defining the "create account"/"manage account" intents already, rather than directly opening a URI?

src/main/java/org/mozilla/gecko/fxa/sync/FxAccount.java
```diff
@@ -0,0 +1,144 @@
...
+  public void login(final Context context, final BrowserIDKeyPair keyPair, final Delegate delegate) {
+    ExtendedJSONObject keyPairObject;
+    try {
+      keyPairObject = new ExtendedJSONObject(keyPair.getPublic().serialize());
+    } catch (Exception e) {
+      delegate.handleError(e);
+      return;
+    }
+
+    FxAccountClient fxAccountClient = new FxAccountClient(idpEndpoint, Executors.newSingleThreadExecutor());
```
rnewman: Blimey, two separate executor threads, and I'm only on line 81!

src/main/java/org/mozilla/gecko/fxa/sync/FxAccount.java
```diff
@@ -0,0 +1,144 @@
...
+    fxAccountClient.sign(sessionTokenBytes, keyPairObject, JSONWebTokenUtils.DEFAULT_CERTIFICATE_DURATION_IN_MILLISECONDS,
+        new FxAccountClient.RequestDelegate<String>() {
```
rnewman: Break out this delegate to a named inner class?

src/main/java/org/mozilla/gecko/fxa/sync/FxAccount.java
```diff
@@ -0,0 +1,144 @@
...
+      public void handleSuccess(String certificate) {
+        Logger.pii(LOG_TAG, "Got certificate " + certificate);
+
+        try {
+          String assertion = JSONWebTokenUtils.createAssertion(keyPair.getPrivate(), certificate, authEndpoint);
+          Logger.pii(LOG_TAG, "Generated assertion " + assertion);
+
+          JSONWebTokenUtils.dumpAssertion(assertion);
+
+          String uri = authEndpoint + (authEndpoint.endsWith("/") ? "" : "/") + "1.0/sync/1.1";
```
rnewman: Lift out this logic.

src/main/java/org/mozilla/gecko/fxa/sync/FxAccountGlobalSession.java
```diff
@@ -0,0 +1,53 @@
...
+import org.mozilla.gecko.sync.net.AuthHeaderProvider;
+import org.mozilla.gecko.sync.stage.CheckPreconditionsStage;
+import org.mozilla.gecko.sync.stage.GlobalSyncStage;
+import org.mozilla.gecko.sync.stage.GlobalSyncStage.Stage;
+
+import android.content.Context;
+import android.os.Bundle;
+
+public class FxAccountGlobalSession extends GlobalSession {
+  private static final String LOG_TAG = FxAccountGlobalSession.class.getSimpleName();
```
rnewman, line 30: It occurs to me that we should define:

```
  Logger:
    public static String getLogTag(Class clazz) {
      return clazz.getSimpleName().substring(...);
    }
```

to make this a little less error-prone in the case of long class names.

rnewman:
Looks like a good starting point. Gonna enjoy working on the perf bits.

src/main/java/org/mozilla/gecko/fxa/sync/FxAccountGlobalSession.java
ncalexan, line 30: I wouldn't be against such a static method for clarity, but `Logger` doesn't have the arbitrary limit on tag lengths.  Follow-up.

src/main/java/org/mozilla/gecko/fxa/sync/FxAccount.java
ncalexan: Damn you!  I already lifted this out a little, and identified it as a smell.  I was thinking about this as I was following asleep and it's very close to the authenticator's "token type".  Will lift.

src/main/java/org/mozilla/gecko/fxa/sync/FxAccount.java
ncalexan: My new style is to give everything an executor, and worry about threading the same executor through where appropriate later.

src/main/java/org/mozilla/gecko/fxa/authenticator/FxAccountAuthenticator.java
ncalexan: Agree.

PR #375, opened 2013-12-31
Bug 955808 - update FxAccountClient to "onepw" protocol

ncalexan:


rnewman:
The only part I'm unhappy about is the perf of the key stretching. We can get a big boost with negligible complexity by using the scrypt code in 138e62563f9b6bfcf151592807bc33cb7f7432b6, so I think it's worth tracking that work as a follow-up (even if we just land compiled libraries in `m-c`, building them in-tree and copying them in `fennec-copy-code.sh`).

I definitely wouldn't want to add eight or more seconds to the login phase for the majority of users, when we can get that down to below a second.

ncalexan:
Pretty hefty update; up to you how much re-review this gets.  This is working locally and against the onepw remote server.

Agreed on perf; we'll drop in NativeCrypto when the schedule forces everyone's hand :)

rnewman:
I think this looks good, but the proof will be in the pudding! r+ with nits.

ncalexan:
Landed.

PR #376, opened 2014-01-01
Bug 915312 - pbkdf2 only

ncalexan:
This goes on top of some m-c patches, to be pushed in a second.


rnewman:
r=me on the two patches I didn't write :D

ncalexan:
This landed in a different form.

PR #380, opened 2014-01-06
Bug 956816 - deserialize keypairs

ncalexan:
Just a step on the path.  Nice easy review.


rnewman:
First two parts trivial r=me.

Last part looks good, but a quick sec-review-needed would be worthwhile before this gets much closer to shipping.

PR #381, opened 2014-01-07
Nalexander/fxa ui

ncalexan:
@rnewman take a look.


rnewman:
Phone comment, capturing for posterity: we need to introduce another state in the account state machine, which will capture the state in which we have a user's password but have not yet logged in (password => token). This is a step along the line from when we have neither (such as when a token has been invalidated).

We'll temporarily enter this state when login fails after account creation, and this will be our entry point after a migration sentinel has been processed.

rnewman:
Nothing in the UI code stands out as evil. So long as it matches the flows and functions, I'm happy. Continue to iterate!

ncalexan:
Let's close this, since it's essentially all landed.

PR #385, opened 2014-01-18
Bug 929066 - Handle skew in FxA login.

rnewman:
Untested, but see what you think.


ncalexan:
Make 7110600 a Pre: commit, and I prefer all the inner commits to list the ticket.  This looks good, r=nalexander for these parts.

rnewman:
Comments addressed, but I'll add some tests before I land this.

ncalexan:
Landed.

PR #393, opened 2014-01-22
Nalexander/bug 962668 assertion skew

ncalexan:


src/main/java/org/mozilla/gecko/tokenserver/TokenServerClient.java
```diff
@@ -203,11 +204,15 @@ public TokenServerToken processResponse(HttpResponse response)
...
   public void getTokenFromBrowserIDAssertion(final String assertion, final boolean conditionsAccepted,
       final TokenServerClientDelegate delegate) {
-    BaseResource r = new BaseResource(uri);
+    final BaseResource r = new BaseResource(uri);
 
     r.delegate = new BaseResourceDelegate(r) {
       @Override
       public void handleHttpResponse(HttpResponse response) {
+        SkewHandler skewHandler = SkewHandler.getSkewHandlerForResource(r);
+        if (skewHandler != null) {
```
rnewman: This should never be null, but that you're checking implies that you'd like it to be…?

src/main/java/org/mozilla/gecko/tokenserver/TokenServerClient.java
```diff
@@ -203,11 +204,15 @@ public TokenServerToken processResponse(HttpResponse response)
...
+          skewHandler.updateSkew(response, System.currentTimeMillis());
```
rnewman: Ideally the wall clock time should be computed as early as possible, so lift `now` up to the start of the method.

(Example: what if fetching the skew handler causes a GC that takes 800ms?)

src/main/java/org/mozilla/gecko/tokenserver/TokenServerClient.java
ncalexan: Nah, just being safe.  I checked before part 2 and see it'll never be null, will update.

src/main/java/org/mozilla/gecko/tokenserver/TokenServerClient.java
ncalexan: Will do.

PR #396, opened 2014-01-24
Bug 963424 - Use token server-provided cluster URL 

rnewman:
Passes tests.


ncalexan:
Can we drop the Part 1 commit entirely?

rnewman:
Merged.

PR #398, opened 2014-01-25
Nalexander/bad case

ncalexan:


rnewman:
https://i.chzbgr.com/maxW500/3503900160/h2AAF2EF3/

rnewman:
Otherwise, looks good.

PR #399, opened 2014-01-25
Bug 959774 - TOS links for FxA.

rnewman:


ncalexan:
According to pike and flod, it's best not to embed HTML in DTDs at all.  So we need to build the string and the href's manually out of lots of little strings, and then instruct translators about the little strings.  But `openLinkInFennec` is generally useful; let's have that in some more general place.  It might even go in `BrowserContract` or something else provided by Fennec.

rnewman:
Thoughts?

ncalexan:
lgtm, but I'd like it if we can fold this into the original two commits.

ncalexan:
Oh, can you grow this for all the links?  I see fxaccount_account_create_not_allowed_learn_more and fxaccount.old.firefox.  Pretty sure they'll need the linkify call updated.

ncalexan:
Hitting https://bugzilla.mozilla.org/show_bug.cgi?id=962542 and https://bugzilla.mozilla.org/show_bug.cgi?id=961170 would be nice, too.  Both go to some URL on the web (we're not going to handle forgot password locally for Fx 29).  ckarlof and nchapman would know the URLs.

ncalexan:
SHIP IT

rnewman:
Gotta solve https://bugzilla.mozilla.org/show_bug.cgi?id=964883 first. Might involve us always starting Sync setup in a new task — i.e., using the task stack in place of the back stack.

rnewman:
This landed. SUMO link work is in #402.

PR #405, opened 2014-01-29
Bug 959915 - Adapt FxA token server client to include X-Client-State header.

rnewman:


rnewman:
This test vector hasn't been compared against desktop yet, due to an error in the desktop patch.

PR #406, opened 2014-01-30
Nalexander/bug 965367 choose what to sync

ncalexan:
This can't be tested against prod yet because of issues with the production 1.5 Sync endpoints.  It's definitely writing the prefs, and definitely the creation logic worked against dev endpoints, so let's review and we'll fix issues as we land tomorrow.


rnewman:
Merged.

PR #417, opened 2014-02-06
Bug 958900 - Schedule a periodic sync for new FxAccountSyncAdapter.

rnewman:
WIP.


ncalexan:
Can you fold Bug 958900 down into coherent commits?  There's a lot of inter-commit churn that makes it hard to follow the thread.

ncalexan:
OK, this seems reasonable.  I have some niggles, and a confusion about this actually working, but I have faith :)  r=me

PR #421, opened 2014-02-26
Nalexander/bug 974134 status

ncalexan:
This is working well locally.  The weird code ordering is due to headers in Part 1 (which I didn't want to change); I'll put the final pre as a post, I think.

I have not tested on devices older than 3.0.  I have a 2.3.6 armv6 phone, but it's a pain to build for.  I think it makes sense to land, and then address old phones with QA, rather than block landing on a try run and test cycle, but I could be convinced otherwise.

This can be uplifted.  Already opened a ticket for action bar/up support (which can be uplifted), and will open a new ticket for the "Signed in as" string.


rnewman:
Rubberstamp!

ncalexan:
Closed via https://github.com/mozilla-services/android-sync/commit/d4389e525b5aa477afcabcf2b7c4524c066d1e8c

PR #426, opened 2014-03-04
Nalexander/bug 979075 http head for token server skew

ncalexan:
Probably not necessary now, but I did it 'cuz I was tired of seeing failures during testing.  We should land it anyways.

There's no dep on earlier work, just lazy tree management.


ncalexan:
Closing this because earlier pieces have landed, and we don't need Bug 979075 right now.  We can re-visit if necessary.

PR #427, opened 2014-03-04
Bug 966143 - Don't attempt to process a malformed password record.

rnewman:
We don't have any sanity checking, and neither does PasswordsProvider, apparently. The SQL exception causes the separate process to crash, which is caught by the Android crash reporter. Oops.


ncalexan:
lgtm.

PR #429, opened 2014-03-05
Bug 957894: Add fxa account pickling code.

mcomella:


rnewman:
This won't let me add line-by-line comments…

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
```diff
@@ -0,0 +1,273 @@
...
+import org.mozilla.gecko.fxa.login.State.StateLabel;
+import org.mozilla.gecko.fxa.login.StateFactory;
+import org.mozilla.gecko.sync.ExtendedJSONObject;
+import org.mozilla.gecko.sync.NonObjectJSONException;
+import org.mozilla.gecko.sync.Utils;
+
+import android.accounts.Account;
+import android.content.Context;
+
+// TODO: Reword this comment, or point a finger at sync's AccountPickler.
```
rnewman: If the two `AccountPickler`s implemented the same interface, you could inherit the comment.

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
```diff
@@ -0,0 +1,273 @@
...
+ * ACTION_EXTERNAL_APPLICATIONS_AVAILABLE broadcast Intent, which will notify
+ * your application when applications installed on external storage have become
+ * available to the system again. At which time, you can restart your Service.
+ * <p>
+ * Problem: <a href="http://code.google.com/p/android/issues/detail?id=8485">that intent doesn't work</a>!
+ */
+public class AccountPickler {
+  public static final String LOG_TAG = "AccountPickler";
+
+  // TODO: What to do with version?
```
rnewman: What do you mean by this comment? You use it below…

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
```diff
@@ -0,0 +1,273 @@
...
+    o.put(KEY_ACCOUNT_TYPE, FxAccountConstants.ACCOUNT_TYPE);
+    o.put(KEY_EMAIL, account.getEmail());
+    o.put(KEY_PROFILE, account.getProfile());
+    o.put(KEY_IDP_SERVER_URI, account.getAccountServerURI());
+    o.put(KEY_TOKEN_SERVER_URI, account.getTokenServerURI());
+    o.put(KEY_IS_SYNCING_ENABLED, account.isSyncingEnabled());
+
+    // TODO: If prefs version changes under us, SyncPrefsPath will change, "clearing" prefs.
+
+    // TODO: In the future, this should migrate for us, rather than returning null, right?
```
rnewman: Yes.

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
```diff
@@ -0,0 +1,273 @@
...
+   *
+   * @param AndroidFxAccount the account to persist to disk
+   * @param filename name of file to persist to; must not contain path separators.
+   */
+  public static void pickle(final AndroidFxAccount account, final String filename) {
+    final ExtendedJSONObject o = new ExtendedJSONObject();
+    o.put(KEY_PICKLE_VERSION, new Long(PICKLE_VERSION));
+    o.put(KEY_PICKLE_TIMESTAMP, new Long(System.currentTimeMillis()));
+
+    o.put(KEY_ACCOUNT_VERSION, AndroidFxAccount.CURRENT_PREFS_VERSION);
```
rnewman: I think you mean `CURRENT_ACCOUNT_VERSION`. `CURRENT_PREFS_VERSION` should be irrelevant to the pickling code — it doesn't need to read prefs. And see below…

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
```diff
@@ -0,0 +1,273 @@
...
+    o.put(KEY_ACCOUNT_TYPE, FxAccountConstants.ACCOUNT_TYPE);
+    o.put(KEY_EMAIL, account.getEmail());
+    o.put(KEY_PROFILE, account.getProfile());
+    o.put(KEY_IDP_SERVER_URI, account.getAccountServerURI());
+    o.put(KEY_TOKEN_SERVER_URI, account.getTokenServerURI());
+    o.put(KEY_IS_SYNCING_ENABLED, account.isSyncingEnabled());
+
+    // TODO: If prefs version changes under us, SyncPrefsPath will change, "clearing" prefs.
```
rnewman, line 102: I think what you mean is “if we pickled an account which is expecting one prefs version, and we unpickle after `CURRENT_PREFS_VERSION` increments, then `AndroidFxAccounts` will need to handle that in exactly the same way as if the package were upgraded”.

Also, this case shouldn't happen: you can't upgrade an Android package that's not on a mounted filesystem.

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
rnewman: Note, though, that this illustrates that accounts don't hold on to their current pref version.

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
```diff
@@ -0,0 +1,273 @@
...
+      ps = new PrintStream(fos);
+      ps.print(pickle.toJSONString());
+      Logger.debug(LOG_TAG, "Persisted " + pickle.keySet().size() + " account settings to " +
+          filename + ".");
+    } catch (Exception e) {
+      Logger.warn(LOG_TAG, "Caught exception persisting account settings to " + filename +
+          "; ignoring.", e);
+    } finally {
+      if (ps != null) {
+        ps.close();
```
rnewman: This is incorrect. If `ps` is null, you'd still need to close `fos`. You need to rearrange the `try` block to acknowledge the existence of two resources — nesting is the usual approach.

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
```diff
@@ -0,0 +1,273 @@
...
+   * Create Android account from saved JSON object.
+   *
+   * @param context
+   *          Android context.
+   * @param filename
+   *          name of file to read from; must not contain path separators.
+   * @return created Android account, or null on error.
+   */
+  public static AndroidFxAccount unpickle(final Context context, final String filename) {
+    // TODO: Assume this will be called only when account DNE? imo, it's clearer.
```
rnewman: The Sync 1.1 version actually only calls this from within `syncAccountsExist`, so you definitely know that it doesn't! Feel free to document this constraint in the Javadoc.

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
```diff
@@ -0,0 +1,273 @@
...
+      return null;
+    }
+
+    final AndroidFxAccount account;
+    try {
+      // TODO: Not the same context as when it was created. Okay?
+      account = AndroidFxAccount.addAndroidAccount(context, params.email, params.profile,
+          params.idpServerURI, params.tokenServerURI, params.state, params.isSyncingEnabled, true);
+    } catch (Exception e) {
+      // TODO: Catch all okay?
```
rnewman: Yeah.

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
```diff
@@ -0,0 +1,273 @@
...
+      Logger.warn(LOG_TAG, "Exception when adding Android Account; aborting.", e);
+      return null;
+    }
+
+    if (account == null) {
+      Logger.warn(LOG_TAG, "Failed to add Android Account; aborting.");
+      return null;
+    }
+    account.persistBundle(params.bundle);
```
rnewman: Perhaps have `AndroidFxAccount` have an adder that takes a bundle?

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
```diff
@@ -0,0 +1,273 @@
...
+        throws NonObjectJSONException, NoSuchAlgorithmException, InvalidKeySpecException {
+      // Sanity check.
+      final String accountType = json.getString(KEY_ACCOUNT_TYPE);
+      if (!FxAccountConstants.ACCOUNT_TYPE.equals(accountType)) {
+        throw new IllegalStateException("Account type has changed from, " + accountType +
+            ", to, " + FxAccountConstants.ACCOUNT_TYPE + ".");
+      }
+
+      // TODO: What to do with account version information? Ignore it? We would migrate
+      // based on pickle version by the time we return from fromJSON anyway.
```
rnewman: The account itself will care.

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
```diff
@@ -0,0 +1,273 @@
...
+      this.idpServerURI = json.getString(KEY_IDP_SERVER_URI);
+      this.tokenServerURI = json.getString(KEY_TOKEN_SERVER_URI);
+      this.isSyncingEnabled = json.getBoolean(KEY_IS_SYNCING_ENABLED);
+
+      this.bundle = (ExtendedJSONObject) json.getObject(KEY_BUNDLE);
+      if (bundle == null) {
+        throw new IllegalStateException("Pickle bundle is null.");
+      }
+      // TODO: Should copy-pasta BUNDLE_KEY_STATE & LABEL to this file to ensure we maintain
+      // old versions?
```
rnewman, line 257: If you don't include them, the account will throw exceptions when accessed…

src/main/java/org/mozilla/gecko/fxa/authenticator/AndroidFxAccount.java
```diff
@@ -41,6 +41,7 @@
 
   public static final int CURRENT_PREFS_VERSION = 1;
 
+  // TODO: Add comment that if you add fields, update pickle.
```
rnewman: …

src/main/java/org/mozilla/gecko/fxa/authenticator/AndroidFxAccount.java
```diff
@@ -82,6 +83,11 @@ public AndroidFxAccount(Context applicationContext, Account account) {
     this.accountManager = AccountManager.get(this.context);
   }
 
+  // TODO: Docs. This vs. Bundle.
```
rnewman: …

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
mcomella, line 257: I was specifically referring to the keys' values. If the constants' values are changed in `AndroidFxAccount`, our pickled data will not be restored correctly. Do you think they should be copied over?

src/main/java/org/mozilla/gecko/fxa/authenticator/AndroidFxAccount.java
mcomella: While I suppose it's obvious, if someone is unaware there's a pickler, maybe they won't update it. Better to be explicit, right?

src/main/java/org/mozilla/gecko/fxa/authenticator/AndroidFxAccount.java
mcomella: From someone approaching this without fxa context, the difference between bundling and pickling might be unclear, no? It's easy to skip over the fact that an account contains a bundle, and the method names (e.g. `account.unbundle`) imply you're unbundling the account, not the bundle within the account.

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
mcomella: Good idea, but this is kind of messy because the `pickle` methods do not have the same method signature and leaving out that essential method from the interface is a bit fishy. I could try to mangle both methods to the same signature, but it would be messy and require changing a lot of already working (and soon to be antiquidated) code. 

Any ideas?

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
mcomella: I shove it into the file during pickle, but I never unpickle it, verify the version, etc. I suppose this will be taken care of when migration is implemented though.

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
mcomella: 7ec9077fd112a2006d946e5b645bbf156b80d484

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
mcomella: Removed comment: f71b4b8ebcd3457422e89193ef2fd806baf98eed

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
mcomella: Removed comment: f71b4b8ebcd3457422e89193ef2fd806baf98eed

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
mcomella: I really need to scrutinize code I copy-paste more thoroughly. x_x 2dcfb1797a479a9da6a814b0b2e78eaf61169f78

Should I correct the sync implementation too?

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
mcomella: 34b8968ce1a8b5faa9214ecb62cf6b511a782052

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
mcomella: Comment removed: 2e038c37151e33dc65adbe338d2ac3588645e5c2

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
mcomella: What do you mean? We can only create an account [of the current version](https://mxr.mozilla.org/mozilla-central/source/mobile/android/base/fxa/authenticator/AndroidFxAccount.java#294) so we have to upgrade the pickle data before we create an account with it.

This means we can avoid storing the account version at all, and just upgrade based on pickle version, assuming we upgrade the pickle version each time we upgrade the account version.

src/main/java/org/mozilla/gecko/fxa/authenticator/AndroidFxAccount.java
mcomella: Let me know if you want me to change it: 9930719fa8e9ff5ef8e2e197f85c0db2fbc06fd3

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
rnewman: 0fg

src/main/java/org/mozilla/gecko/fxa/authenticator/AndroidFxAccount.java
rnewman: I mean "write the docs". You just did :)

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
rnewman, line 257: Can you just refer to them in AndroidFxAccount? Package scope?

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
rnewman: Provide a mechanism to create an account with an older version (just raw data in), and let the account code take care of the upgrade…?

rnewman:
If it works, land it.

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
mcomella, line 257: Do you mean refer to them in `AccountPickler`? I can, and this compiles, but if the values are ever changed in a later revision, the pickled files on disk would not be readable unless we remember to include the old value in `AccountPickler`. Perhaps I'm being overly cautious.

src/main/java/org/mozilla/gecko/fxa/authenticator/AndroidFxAccount.java
mcomella: It was a bit messy to do in javadoc, but I tried: 9d03582d17c5b29d80687896a41f01dc84f24643

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
mcomella: Left a migration TODO: 2fb6e8f68e4c8ffdf93755d7e32bc03479c6b0c6

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
mcomella: 47ef757579c2c5edf6ef0f76cec0d950524f5230

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
mcomella: Sure, then I'll leave in the TODO.

src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
mcomella, line 102: What I meant with the TODO is: We're relying on the Preferences remaining between `pickle` and `unpickle`, but we use a different path if the version changes, meaning the preferences will be empty if `CURRENT_PREFS_VERSION` is incremented.

I think it's a misunderstanding (if so, ignore below), but otherwise I'll try to respond directly:

> Also, this case shouldn't happen: you can't upgrade an Android package that's not on a mounted filesystem.

This case being a package being upgraded? I suppose your statements are true then, however, I think it's irrelevant to the preference version being incremented.

mcomella:
> If it works, land it.

Unknown. Test time. Note that my changes today have been slightly rushed and less thorough than I'm accustomed to because this comment implies we need to make merge.

Also, TODO: put pickle/unpickle in the actual code.

mcomella:
Note that part 3 & 4 have no automated tests, nor have been tested locally - I'm working on the latter now.

mcomella:
Compiles and tests still work locally - I'm not going to retest by hand.

mcomella:
Actually, that's just lazy, working on it. :P

I'll rebase first though.

mcomella:
github doesn't automagically these closed anymore, it seems :sweat:: 88719342e9b462f4d0161aac0563acfaeb86d8e1

PR #431, opened 2014-03-08
Bug 969672 - Handle declined engines in meta/global.

rnewman:
As far as I can tell, this successfully:
1. Records declined engines during setup.
2. Persists declined engines encountered via meta/global. (Note that we don't merge these with the local choices. This shouldn't be an issue unless there's a race between two clients declining engines at the same time!)
3. Uploads the same.

Tests are no worse than they were before, and I fixed one that was NPE-failing due to variable shadowing.

This should be enough to support desktop use-cases.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/GlobalSession.java
+++ b/src/main/java/org/mozilla/gecko/sync/GlobalSession.java
@@ -18,6 +18,7 @@ import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicLong;
 
+import org.json.simple.JSONArray;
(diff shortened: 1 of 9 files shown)


rnewman:
New review please, Nick!

ncalexan:
This is looking really good (I found it much easier to follow Part 2 through 4).  Part 2 seems mis-named, since the new function has no callers.

I see no special handling of the fact that the History checkbox means both the "history" and the "forms" engine.  It appears that you always piggy-back on the existing special handling for user selected engines, but please verify that this is doing the right thing in this case.

Looks good to me!

PR #436, opened 2014-03-14
Nalexander/bug 974124 borders padding colours

ncalexan:
No sense doing a bunch of tiny PRs for these things.  First 2 will want uplift; last 3 are strings.


rnewman:
Merged and landed.

PR #440, opened 2014-03-25
Bug 983856 - Message when syncing is disabled by Android.

ncalexan:
This shows two new "error" (really, warning) states.  Both are shown
when the Firefox Account is correctly configured, and ready to sync, but
Android is preventing syncing automatically for some reason.  The first
is when Android is not syncing _any_ Account
automatically (getMasterSyncAutomatically is false); the second is when
Android is not syncing *this Account
automatically (getSyncAutomatically) is false.


ncalexan:
Landed.  This required a rework of @mcomella's freshly added `AndroidFxAccount.isSyncEnabled`, which was sensible but not correct.  Integration tests passed, save for an unrelated NPE in testDeclinedEngines.

PR #443, opened 2014-03-27
Bug 989105 - Redirect Send Tab to Status activity when the Firefox Accou...

ncalexan:
...nt definitely cannot send a tab.

### DIFF
--- a/src/main/java/org/mozilla/gecko/sync/setup/activities/SendTabActivity.java
+++ b/src/main/java/org/mozilla/gecko/sync/setup/activities/SendTabActivity.java
@@ -14,7 +14,9 @@ import org.mozilla.gecko.R;
 import org.mozilla.gecko.background.common.log.Logger;
 import org.mozilla.gecko.fxa.FxAccountConstants;
 import org.mozilla.gecko.fxa.activities.FxAccountGetStartedActivity;
+import org.mozilla.gecko.fxa.activities.FxAccountStatusActivity;
 import org.mozilla.gecko.fxa.authenticator.AndroidFxAccount;
+import org.mozilla.gecko.fxa.login.State.Action;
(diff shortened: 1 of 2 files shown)


src/main/java/org/mozilla/gecko/sync/setup/activities/SendTabActivity.java
```diff
@@ -224,7 +226,20 @@ public void onResume() {
 
     final Account[] fxAccounts = accountManager.getAccountsByType(FxAccountConstants.ACCOUNT_TYPE);
     if (fxAccounts.length > 0) {
-      this.tabSender = new FxAccountTabSender(applicationContext, fxAccounts[0]);
+      final AndroidFxAccount fxAccount = new AndroidFxAccount(applicationContext, fxAccounts[0]);
+      if (fxAccount.getState().getNeededAction() != Action.None) {
+        // We have a Firefox Account, but it's definitely not able to send a tab
+        // right now. Redirect to the status activity.
+        Logger.warn(LOG_TAG, "Firefox Account named like " + Utils.obfuscateEmail(fxAccounts[0].name) +
```
rnewman: We can't get the obfuscated email from the `AndroidFxAccount`? Worth adding that?

src/main/java/org/mozilla/gecko/sync/setup/activities/SendTabActivity.java
```diff
@@ -224,7 +226,20 @@ public void onResume() {
...
+            " needs action before it can send a tab; redirecting to status activity.");
+        final Intent intent = new Intent(applicationContext, FxAccountGetStartedActivity.class);
```
rnewman: This stanza is already done elsewhere; extract a method.

rnewman:
r+ with changes.

PR #453, opened 2014-04-16
Bug 962622 - rounded corners

ncalexan:
Couple of tickets to look at here, @rnewman.


ncalexan:
All landed a long time ago.

PR #457, opened 2014-05-14
Bug 958889 - Add getFirefoxAccountState method.

mcomella:


ncalexan:
I'm okay with these, modulo requests, but I think it makes more sense to have the caller provide a non-null Android `Account` object, to be slightly more efficient.

ncalexan:
I'll take your word that the refactorings are correct (or better yet, automated).  Fold the review comments into the appropriate patches, please, and ship it!

PR #460, opened 2014-05-29
Bug 788688 - set client device name

ncalexan:
This ended up being more complicated than expected.  Pay particular attention to the Preferences hi-jinks; it's delicate.  I could see not worrying about this corner case if it caused problems in testing.  Observe also the less-than-general approach to force a clients and tabs upload on client name change.  An alternate route would be to invalidate timestamps, but that's really hacky when done from the Status activity.


rnewman:
LGTM, modulo testing on tablet etc.

ncalexan:
Merged.

PR #464, opened 2014-06-10
Bug 1022749 - show custom servers

ncalexan:
And the UI for exposing this, read-only, during setup and as part of the status activity.


ncalexan:
Landed.

PR #465, opened 2014-06-16
Bug 1026005 - accounts loader

ncalexan:
This builds on top of some test only fixes, and hasn't had a whole lot of testing.  I'll be building additional Remote Tabs stuff on top of it, which will fill in the testing gaps.

### DIFF
--- a/android-sync-instrumentation/pom.xml
+++ b/android-sync-instrumentation/pom.xml
@@ -65,6 +65,13 @@
       <scope>provided</scope>
       <type>jar</type>
     </dependency>
+
+    <dependency>
+      <groupId>com.android.support</groupId>
+      <artifactId>support-v4</artifactId>
+      <version>18.0.0</version>
+      <scope>provided</scope>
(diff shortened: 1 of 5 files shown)


mcomella:
Ignoring parts 1-5 and everything before and in-between as I've already reviewed that code. Why hasn't it landed yet?

edit: lol, I'm stupid. Ignore me. :)

ncalexan:
Argh, pushed a rebase version and github hides the comments.  They're in your email, and I went with a separate "review comments" commit for simplicity.  We can rebase the new work on the old and repush, which should rehabilitate our messages, if we need to.

mcomella:
> I don't really understand the confusion with shadowing, but I'm going to rename the variables in loadInBackground as you suggest.

I generally find shadowing to be difficult to read because if you ever forget that the current method redefined the variable, things can get confusing. Also, it can create confusion when writing/modifying the code if the local declaration is removed, particularly accidentally.

PR #466, opened 2014-06-16
Bug 1025937 - Silently drop null intents in HealthReportBroadcastService.

rnewman:
It looks like an IntentService can sometimes be restarted with a null intent. The right thing to do appears to be to ignore it — we'll get another intent if we have to do some work. (If we were restarted mid-operation, we just wait until the next time.)

### DIFF
--- a/src/main/java/org/mozilla/gecko/background/healthreport/HealthReportBroadcastService.java
+++ b/src/main/java/org/mozilla/gecko/background/healthreport/HealthReportBroadcastService.java
@@ -100,6 +100,11 @@ public class HealthReportBroadcastService extends BackgroundService {
   protected void onHandleIntent(Intent intent) {
     Logger.setThreadLogTag(HealthReportConstants.GLOBAL_LOG_TAG);
 
+    // Intent can be null. Bug 1025937.
+    if (intent == null) {
+        Logger.debug(LOG_TAG, "Short-circuiting on null intent.");
+    }
+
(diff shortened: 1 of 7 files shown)


mcomella:
There appears to be `onHandleIntent` calls in `fxa.receivers.FxAccountDeletedService` and `sync.receivers.SyncAccountDeletedService` - is there any reason you didn't add the fix there as well?

src/main/java/org/mozilla/gecko/background/healthreport/HealthReportBroadcastService.java
```diff
@@ -113,6 +118,11 @@ protected void onHandleIntent(Intent intent) {
    * Attempts to handle the given intent for FHR document upload. If it cannot, false is returned.
    */
   protected boolean attemptHandleIntentForUpload(final Intent intent) {
+    if (intent == null) {
```
mcomella: Why did you only do this for `HealthReportBroadcastService`? `AnnouncementsBroadcastService` uses a similar paradigm.

rnewman: I spent the most time in that file.

Updated push coming momentarily.

rnewman:
This original push only touched descendents of `BackgroundService`, which the deleted services are not. Updated push coming momentarily.

rnewman:
Comments addressed and merged. Thanks!

PR #468, opened 2014-06-26
Bug 1030740 - Add return statements inadvertently omitted from Bug 1025937.

rnewman:
Oops. Not much of a short-circuit without the return, is it?


mcomella:
r+ - lgtm.

PR #473, opened 2014-07-28
Bug 1045085 - Remove Product Announcements.

rnewman:


rnewman:
Note that we don't yet clean up stale prefs.

PR #475, opened 2014-08-18
Bug 958341 - Present "motivating service" to /account/create calls.

ncalexan:
This is used for metrics and for personalizing the welcome emails.
Since no email gets sent for /account/login, it is not needed there.

### DIFF
--- a/src/main/java/org/mozilla/gecko/background/fxa/FxAccountClient20.java
+++ b/src/main/java/org/mozilla/gecko/background/fxa/FxAccountClient20.java
@@ -104,20 +104,15 @@ public class FxAccountClient20 extends FxAccountClient10 implements FxAccountCli
    * present in our JUnit 4 tests.
    */
   protected URI getCreateAccountURI(final boolean getKeys, final String service) throws UnsupportedEncodingException, URISyntaxException {
+    if (service == null) {
+      throw new IllegalArgumentException("service must not be null");
+    }
     final StringBuilder sb = new StringBuilder(serverURI); // serverURI always has a trailing slash.
-    sb.append("account/create");
-    if (getKeys || service != null) {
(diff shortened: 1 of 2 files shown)


ncalexan:
Lightly tested with local unit tests.  No device testing yet, but I doubt there will be issues 'cuz I caught several with unit tests.

src/main/java/org/mozilla/gecko/background/fxa/FxAccountClient20.java
```diff
@@ -94,13 +97,40 @@ public void handleSuccess(int status, HttpResponse response, ExtendedJSONObject
...
+  protected URI getCreateAccountURI(final boolean getKeys, final String service) throws UnsupportedEncodingException, URISyntaxException {
+    final StringBuilder sb = new StringBuilder(serverURI); // serverURI always has a trailing slash.
+    sb.append("account/create");
+    if (getKeys || service != null) {
+      // Be very careful that query parameters are encoded correctly!
+      sb.append("?");
+      if (getKeys) {
+        sb.append("keys=true");
+      }
+      if (service != null) {
```
rnewman: Can we just mandate that the service is never null, and simplify this? Even allowing the empty string should be fine, no?

rnewman:
LGTM, but I don't see a test that verifies that the right parameter exists in the final URL. Did I miss something?

src/main/java/org/mozilla/gecko/background/fxa/FxAccountClient20.java
ncalexan: Yeah, I'll do this.  Empty strings should be fine, I think.

ncalexan:
I made `service` mandatory, and I added a test.

PR #477, opened 2014-08-20
bug-984503 Added menu and dialog to remove firefox account. Implemented ...

Balakrishnan-Vivek:
...account removal logic


ncalexan:
Merged.

PR #480, opened 2014-08-28
Nalexander/bug 984503 remove account

ncalexan:
The first commit from @Balakrishnan-Vivek has my r+.  @rnewman, I'd like your stamp on parts 2-4.  The reason for having a menu is to keep the "Remove Account" button out of sight for most folks.


ncalexan:
Merged.

PR #488, opened 2014-10-11
Bug 1081401: Remove unnecessary autoboxing.

ChrisKitching:
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1081401


ChrisKitching:
Oddly, the patch that worked for this part on fx-team doesn't apply cleanly to this repository, so my fancy script to do this automagically isn't successful.

I'll apply it manually and try again, then.

_sigh_
Such timewasting. Wow.

PR #500, opened 2014-10-13
Bug 966103 : Add a 'Sync now' button to the main FxA status screen

Balakrishnan-Vivek:
- Added a new preference for syn now
- Handled click and states/animation during sync
- Pending : Need to change the icon with a darker theme


ncalexan:
OK, this is looking really quite good.  I'd like to do a few things to land this:
- fold existing commits down into one and rebase onto develop;
- address minor changes as listed above;
- file a separate ticket for adding the icon and the code to animate the spinning icon.  On my devices it doesn't look great; I think the Sync now/Syncing... text will look just fine.
- file another ticket to have the "last synced" timer update once a minute while the activity is in the foreground.

Thanks for your good work on this and your patience!  So close...

ncalexan:
Merging.

PR #504, opened 2014-10-28
Bug 1080242 - locked account

ncalexan:
@rnewman: over to you for a preliminary review.  This isn't final, because:
- the error code isn't determined
- the `unlock/resend_code` endpoint hasn't been speced;
- copy isn't agreed to.

But the ideas are in place.

### DIFF
--- a/src/main/java/org/mozilla/gecko/background/fxa/FxAccountClient10.java
+++ b/src/main/java/org/mozilla/gecko/background/fxa/FxAccountClient10.java
@@ -770,4 +770,46 @@ public class FxAccountClient10 {
     };
     post(resource, new JSONObject(), delegate);
   }
+
+  /**
+   * Request a fresh unlock code be sent to the account email.
+   * <p>
+   * Since the account can be locked before the device can connect to it, the
+   * only reasonable identifier is the account email. Since the account is
+   * locked out, this request is un-authenticated.
+   *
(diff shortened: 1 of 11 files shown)


src/main/java/org/mozilla/gecko/background/fxa/FxAccountClient10.java
```diff
@@ -770,4 +770,46 @@ public void handleSuccess(int status, HttpResponse response, ExtendedJSONObject
...
+   *          identifying account.
+   * @param delegate
+   *          to invoke callbacks.
+   */
+  @SuppressWarnings("unchecked")
+  public void resendUnlockCode(final byte[] emailUTF8, final RequestDelegate<Void> delegate) {
+    final BaseResource resource;
+    final JSONObject body = new JSONObject();
+    try {
+      resource = new BaseResource(new URI(serverURI + "unlock/resend_code"));
```
rfk, line 791: My read of https://github.com/mozilla/fxa-auth-server/issues/801#issuecomment-56592929 is that this should be `account/unlock/resend_code`.

ncalexan, line 791: aye, well spotted.

ncalexan:
This landed.

PR #509, opened 2014-11-25
Bug 1098667 - migrated from sync11 state

ncalexan:


rnewman:
Nothing stands out here as busted. If it works, and it looks fine, then ship it.

ncalexan:
This landed.

PR #514, opened 2014-12-18
Bug 956444 - handle migration sentinel

ncalexan:
This embraces and extends @rnewman's earlier work.

### DIFF
--- a/src/main/java/org/mozilla/gecko/background/testhelpers/DefaultGlobalSessionCallback.java
+++ b/src/main/java/org/mozilla/gecko/background/testhelpers/DefaultGlobalSessionCallback.java
@@ -40,6 +40,10 @@ public class DefaultGlobalSessionCallback implements GlobalSessionCallback {
   }
 
   @Override
+  public void informMigrated(GlobalSession globalSession) {
+  }
+
+  @Override
   public void handleAborted(GlobalSession globalSession, String reason) {
   }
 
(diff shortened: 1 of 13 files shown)


rnewman:
I honestly don't remember writing any of this, so heh.

Looks good to me. I'd like to see some kind of measurement of this: whether that's telemetry, FHR, or something as simple as a URL parameter or object uploaded to the new Sync account, doesn't matter.

ncalexan:
On Mon, Dec 22, 2014 at 10:47 AM, Richard Newman notifications@github.com
wrote:

> I honestly don't remember writing any of this, so heh.

It was a _long_ time ago.

> Looks good to me. I'd like to see some kind of measurement of this:
> whether that's telemetry, FHR, or something as simple as a URL parameter or
> object uploaded to the new Sync account, doesn't matter.
> 
> That's tricky.  Calling telemetry from background services is kinda-sorta
> uncharted territory: neither of us wrote the telemetry stuff for the Search
> Activity and frankly I don't trust it.  Configuring FHR during a Sync
> requires some machinations that I once worked out (I think?) but were
> non-trivial.  Uploading an object to the new Sync account is do-able, but
> the most obvious one is to include something in the client record, which is
> transient.  I suppose it would be possible to upload to a new _migrated_
> collection.  I wonder if it would make sense to try to use the fxa
> mechanism where we add some tracking data to the `/account/login` call.
> It's a little disjointed and delayed, but it might be possible.

If you can show how to arrange for FHR to be configured correctly during a
(Firefox Account) Sync, that would be best.

PR #515, opened 2014-12-20
Vivek/bug 1058806 add month date field to select magic year

Balakrishnan-Vivek:


ncalexan:
@Balakrishnan-Vivek this is the best android-sync contributor patch I've had the pleasure to review.  I've requested some nits and clarifications, but this is exactly what I had in mind.  I'm going to build and test this locally -- COPPA is dangerous territory -- but this is a big f+ with a Christmas bow on top.

Thanks!

ncalexan:
I folded this down to two commits (age helper and UI), rebased onto develop, and started testing.  I can't easily update the PR with those commits so please do that.  I have some small code style issues but let's leave those until we get the behaviour nailed down.

In testing, I see the following issues:
- the magic year is not detected correctly.  Anything later than "2001" triggered the magic year (which is wrong, it should only trigger in 2002 exactly).  I think the check in `maybeEnableMonthAndDayButtons` is wrong -- it's showing the buttons for all years after and including the magic year, not only the magic year.
- if we cycle from Create -> Sign in -> Create, the magic year buttons are lost.  (This is an issue on desktop, too: see https://github.com/mozilla/fxa-content-server/issues/2035).  I think this means we need to extract the "year setting" code to a helper and make `updateFromIntentExtras` use the helper.
- Desktop shows full month names instead of 3 character codes.  I assume this is `Calendar.LONG` in the `getDisplayNames` call.  I bet you did this intentionally, 'cuz we don't have much space.  Can you investigate using the short name in the control and full names in the list?  Or seeing what long names look like in the control?

In the code, let's isolate "is this the magic year" test, extract it to the age helper, and make sure we're testing it thoroughly.  Plus/minus 1 day, month, and year for both the "is magic year" and the "passes age check" logic.

Visually, it looks really good -- nice spacing.  Since the user can't tap days before tapping the month, can we make the days look disabled until the month is entered?  (This might be more effort than it's worth.)

Great work!

ncalexan:
Hey, I've pulled this patch queue and am going to rework them.  I really appreciate the effort you've put in to split them, but the queue isn't quite achieving what we want.  What we want is that each patch is logically independent, meaning that I can read one, mentally "commit" it, and move on.  With the queue here, the logic is tweaked and changed throughout.  I'm going to dig in and try to produce a queue that has the separation of concerns that I'm talking about, and we can discuss that queue further.

ncalexan:
Landed!

PR #517, opened 2014-12-23
Bug 1081389: Stop overriding abstract methods with more abstract methods. r=rnewman

ChrisKitching:


rnewman:
Please fix nit and rebase on top of develop.

PR #523, opened 2015-01-15
Bug 1112725 - Avoid blue line on Next button in Create Account activity.

ncalexan:
Something changed between Android 4.3 and Android 4.4 which folded the
RelativeLayout's marginBottom into the child calculations in some way.
This simple work-around moves the spacing out of the RelativeLayout and
into the marginTop of the control underneath.  All other examples of
this pattern use the marginTop of the control underneath.

Test plan: manually tested on my
- 4.3 Samsung Galaxy S4; and
- 4.4 Google Nexus 5.


ncalexan:
Landed.

PR #527, opened 2015-01-27
Bug 1117830 - Reading list service

rnewman:


rnewman:
Depends on #531. Let's get that landed first.

rfk:
> - final String scope = "profile"; // Niiiiiickkk!

You could consider using scope "profile readinglist" to prepare for the readinglist server switching to a dedicated scope before production release.

rnewman:
Landed in #539.

PR #530, opened 2015-02-10
Bug 1123116

ncalexan:


rnewman:
This looks fine. My main concern is about what happens when the RL build flag is turned off (which it is now, and might be again).

Does that break users who set up FxA before we turn it on? They'll already have migrated to v3.

And does this experience make sense if we land it now?

The obvious way forward, then, is to wrap parts of this code in the build flag. Not ideal.

ncalexan:
Landed.

PR #531, opened 2015-02-10
Bug 1131421 - Initial reading list service work.

rnewman:


rnewman:
I merged `getReadingListPrefs` into Part 4.

PR #532, opened 2015-02-10
Bug 964854 - cache account bundles

ncalexan:
Such guesses, many work-arounds.

### DIFF
--- a/src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
+++ b/src/main/java/org/mozilla/gecko/fxa/authenticator/AccountPickler.java
@@ -278,18 +278,17 @@ public class AccountPickler {
             NonObjectJSONException, NoSuchAlgorithmException {
       // TODO: Should copy-pasta BUNDLE_KEY_STATE & LABEL to this file to ensure we maintain
       // old versions?
-      final StateLabel stateLabel = StateLabel.valueOf(
+      final StateLabel stateLabelString = StateLabel.valueOf(
           bundle.getString(AndroidFxAccount.BUNDLE_KEY_STATE_LABEL));
       final String stateString = bundle.getString(AndroidFxAccount.BUNDLE_KEY_STATE);
-      if (stateLabel == null) {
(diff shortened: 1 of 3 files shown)


rnewman:
This all looks valid, but I think it leaves the largest and riskiest hole.

That is: unless the `setUserData` calls were racing (in which case anything else is unnecessary), we're simply hiding the problem with a memory cache. As soon as our process is killed, we'll hit the same symptoms we get now.

That might be worth having, but perhaps it's just needless complexity.

Are you planning a follow-up to figure out what to do when we get a (partially) empty bundle back, rather than just crashing? E.g., see what we do know, and transition to the appropriate state to re-capture credentials?

ncalexan:
On Tue, Feb 10, 2015 at 4:29 PM, Richard Newman notifications@github.com
wrote:

> This all looks valid, but I think it leaves the largest and riskiest hole.
> 
> That is: unless the setUserData calls were racing (in which case anything
> else is unnecessary), we're simply hiding the problem with a memory cache.
> As soon as our process is killed, we'll hit the same symptoms we get now.
> 
> This is correct, but the point is that we believe _Android's_ in memory
> cache is the issue, and maintaining our own cache avoids hitting
> Android's.  We will still hit Android's cache on process start, but it
> might be the case that Android itself needs to read from disk in order to
> refresh its cache each process start.
> 
> That might be worth having, but perhaps it's just needless complexity.
> 
> Are you planning a follow-up to figure out what to do when we get a
> (partially) empty bundle back, rather than just crashing? E.g., see what we
> do know, and transition to the appropriate state to re-capture credentials?
> 
> This is more work than it seems because the Account state includes things
> that we'd need to create valid account state, namely the server provided
> opaque uid (which we don't actually use).  We could transition some of the
> state details into the Account user data and make it possible to go to
> separated in this case.  (I'm moving to this approach on iOS, where the
> state details are really the state and not the account and the state.)

So my plan is to see if the band-aid helps (up-lifted to beta after a few
days) and re-assess.  I could be plain wrong about Android caching issues :/

ncalexan:
Landed.

PR #533, opened 2015-02-18
Bug 1117829 - oauth and profile clients

ncalexan:


rnewman:
How urgent is it that you land this? Should I wait until the SyncAdapter stuff is done, or split Bug 1117830 and land the first half now, given that the build flag prevents it from being included in Fennec?

PR #534, opened 2015-02-19
Bug 1134454 - Extract common resource code out of Sync.

rnewman:


ncalexan:
This landed.

ncalexan:
Actually, it has not landed yet.  Why not, @rnewman?

rnewman:
Didn't think it had received r+, but I'll assume it now!

PR #547, opened 2015-04-02
Bug 1149226 - Set Reading List sync automatically after upgrade.

ncalexan:
The desired behaviour is: Reading List is syncing automatically if
Firefox Sync is syncing automatically.

Test plan: remove all Firefox Accounts.

1) Install old version.  Add a Firefox Account.  Ensure that Firefox
Sync is enabled.  Upgrade to new version.  Verify that Reading List is
enabled.

2) Install old version.  Add a Firefox Account.  Manually disable
Firefox Sync by unchecking the checkbox in Android Settings >
Accounts.  (This can not be done from the Firefox Account settings
activity; unchecking the 4 boxes there is not the same.)  Upgrade to new
version.  Verify that Reading List is disabled.

3) Install new version.  Add a new Firefox Account.  Verify that Reading
List is enabled.

(There are automated tests for these scenarios.)

### DIFF
--- a/src/main/java/org/mozilla/gecko/fxa/authenticator/AndroidFxAccount.java
+++ b/src/main/java/org/mozilla/gecko/fxa/authenticator/AndroidFxAccount.java
@@ -58,10 +58,15 @@ public class AndroidFxAccount {
   public static final String ACCOUNT_KEY_PROFILE = "profile";
   public static final String ACCOUNT_KEY_IDP_SERVER = "idpServerURI";
 
-  // The audience should always be a prefix of the token server URI.
-  public static final String ACCOUNT_KEY_AUDIENCE = "audience";                 // Sync-specific.
   public static final String ACCOUNT_KEY_TOKEN_SERVER = "tokenServerURI";       // Sync-specific.
   public static final String ACCOUNT_KEY_DESCRIPTOR = "descriptor";
+
+  // The set of authorities to sync automatically changes over time. The first
+  // new authority is the Reading List. This tracks if we've enabled syncing,
+  // and opted in (or out) of syncing automatically, for the new Reading List
+  // authority. This happens either on when the account is created or when
+  // upgrading.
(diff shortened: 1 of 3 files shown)


rnewman:
Looking at the code, it seems like what we want to do (ideally) is:

• On upgrade
• For each `AndroidFxAccount`,
• If `account.getAccountVersion()` call returns `3`, and Sync is enabled, enable RL.
• Bump the current account version so that we don't do that multiple times.

You assert that the approach you've implemented might actually run this migration stuff multiple times, worst-case.

You also assert that the upgrade listener won't run often for release channels.

You don't want to bump the current account version, because doing so is fragile.

Given the above, why not simply do this?

• On upgrade
• For each `AndroidFxAccount`,
• If Sync is enabled, enable RL. (No need to check account version!)
• If all completed, write a pref saying we ran the migration.

It seems like this would achieve the same goal without 80 lines of expected-enabled JSON faffing, no? Just the one runnable that walks accounts, one pref read, one pref write.

ncalexan:
Totally reasonable questions, but:
- It's not safe to use `SharedPrefences` for this, since it gets blown away by (for example) clearing package data in the App manager.  I could be convinced that this isn't worth worrying about.  If we were comfortable saving things in prefs for such sensitive stuff, we should move Account state itself into prefs, avoid the whole Android getUserData bug, and accept that users will have to sign in to their accounts after clearing data in the App manager.  I don't think that users clear App data frequently, but I do think that folks will be surprised if Sync stops working when they do that.
- If we don't do that, we have to seUserData anyways, incurring much of the risk.
- The JSON pfaffing handles this scenario in future.  Perhaps YAGNI, but I expect at least to transition Sync to four checkboxes; and I also think that profile data fits most naturally as a `SyncAdapter` as well.  If we don't do this, we need to remember to do the right thing on upgrade in the future.  (We can, of course, help ourselves by referencing the default authority list in the right place, making this more obvious.)

PR #550, opened 2015-04-08
Bug 1149208 - Make it easier to create a Firefox Account pointing at sta...

ncalexan:
...ge.

Tap the Firefox icon 5 times on the Getting Started screen, and the
account sign up flow will start, pointing at the stage account
constellation.


rnewman:
LGTM.

PR #552, opened 2015-04-14
Bug 1153973 - Don't blindly apply deletions as insertions.

rnewman:
I haven't run this test yet!

### DIFF
--- a/src/main/java/org/mozilla/gecko/reading/LocalReadingListStorage.java
+++ b/src/main/java/org/mozilla/gecko/reading/LocalReadingListStorage.java
@@ -229,7 +229,12 @@ public class LocalReadingListStorage implements ReadingListStorage {
 
     @Override
     public void addDownloadedRecord(ServerReadingListRecord down) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/mozilla/gecko/reading/LocalReadingListStorage.java
```diff
@@ -229,7 +229,11 @@ public void addChangedRecord(ClientReadingListRecord record) {
 
     @Override
     public void addDownloadedRecord(ServerReadingListRecord down) {
-      additionsOrChanges.add(down);
+      final Boolean deleted = down.fields.getBoolean("deleted");
+      if (deleted == null || !deleted.booleanValue()) {
+        additionsOrChanges.add(down);
+      }
+      addDeletion(down.getGUID());
```
ncalexan: Do you want an `else`?  (I'll keep reading.)

ncalexan: Looks like not.

